PR #3, opened 2023-09-21
Add github action to publish lambda layer

pshanmukrao:


pshanmukrao:
Currently I am publishing a manual workflow that can be triggered from UI to publish new lambda layer. We can automate it later depending on the usecase.

sz2376 (dismissed):

stevevls:
@pshanmukrao is this still something we're pursuing?  if so, please rebase to pick up the latest GH status checks.  Otherwise let's close it out.

PR #6, opened 2023-10-11
Remove redundant __subclasshook__ methods, use ABC insted of ABCMeta

stevevls:
The subclass hooks are not required because ABC will prevent instantiation of a class unless all of the abstract methods are implemented.

The change from using ABCMeta meta class to ABC is stylistic in nature only, but it's arguably clearer to a reader who is not familiar with meta classes.


flux_sdk/pension/capabilities/update_deduction_elections/interface.py
```diff
@@ -16,14 +15,6 @@ class UpdateDeductionElections(metaclass=abc.ABCMeta):
...
-    @classmethod
-    def __subclasshook__(cls, subclass):
-        return (
-            (hasattr(subclass, "parse_deductions") and callable(subclass.parse_deductions))
-            or NotImplemented
-        )
-
     @staticmethod
     @abstractmethod
     def parse_deductions(uri: str, stream: IOBase) -> list[EmployeeDeductionSetting]:
```
sz2376, line 20: We had another `parse_deduction(record: dict[str, str])` function defined as below as an alternative to the `parse_deductions(uri: str, stream: IOBase)` function; people only need to implement one of them. How do we plan to handle this use case?

stevevls: That's a good question.  We've made a lot of design decisions since last we visited those ideas such as releasing the SDK as a public library and running the third party code in lambda.

I initially thought the choice between file/row-by-row would be made at scaffolding time and we would emit a customized interface with just a single function.  At module load time, we'd introspect the implementation to determine which of the two functions it implemented and act accordingly.  But now that the SDK is public and fixed, that option won't work.

The lambda decision is important b/c I don't think we'd want to invoke the lambda for each row in the file.  In order to support this feature, then, I think we'd need to have a default implementation of the parse file function that would operate on the lambda side.  So for this class we may decide to add back a custom `__subclass_hook__` to map in the default method in those cases.  WDYT?

Since this is now a public artifact, though, I think we should try to keep speculative stuff work out of it. :)

sz2376 (approved):

PR #18, opened 2023-11-10
Ben Admin Kit

bryanbierce:
- add ben admin scaffold
- update data models


flux_sdk/ben_admin/capabilities/process_deduction_data/interface.py
```diff
@@ -0,0 +1,34 @@
+from abc import ABC, abstractmethod
+
+from flux_sdk.flux_core.data_models import CompanyInfo, File
+from flux_sdk.ben_admin.capabilities.process_deduction_data.data_models import AppConfig
+
+
+class ProcessDeductionData(ABC):
```
stevevls, line 7: Can we rename this to `ProcessDeductions` instead?  I don't think the "Data" tells us about what this capability _does_.

flux_sdk/ben_admin/capabilities/process_deduction_data/interface.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    This class represents the "process deduction data" capability. The developer is supposed to implement
+    the following methods in their implementation. 
+
+    The methods on this class are required to instantiate an instance of this class.
+    """
+
+    @staticmethod
+    @abstractmethod
+    def get_file_name(app_config: AppConfig, company_info: CompanyInfo) -> str:
```
stevevls, line 17: In other kits, we've been moving away from providing a hook like this. It should be solve-able with configuration instead.  Is this strictly required or did you add it based on an example elsewhere?

flux_sdk/ben_admin/capabilities/process_deduction_data/interface.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        This method receives the basic company info and is expected to return the file name which should be retrieved with the deduction info
+
+        :param CompanyInfo:
+        :return str:
+        """
+
+    @staticmethod
+    @abstractmethod
+    def process_deductions_file(deductions_file: File):
```
stevevls, line 27: What sort of processing is happening in this function?  Who is doing the processing and what is the end result?

flux_sdk/ben_admin/capabilities/process_deduction_data/data_models.py
```diff
@@ -0,0 +1,5 @@
+class AppConfig:
```
stevevls, line 1: Can we name this something more descriptive?  I'm not sure what the app is that's getting configured.  You can look at other kits for examples.

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
```diff
@@ -0,0 +1,109 @@
+from datetime import datetime
+from enum import Enum
+from typing import Optional
+
+from flux_sdk.flux_core.data_models import Address, EmployeeState, Gender
+
+
+class AppConfig:
```
stevevls, line 8: Same comment regarding name.  Alternately, is this something that belongs as a kit-level configuration?

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
```diff
@@ -0,0 +1,109 @@
...
+    '''
+    This contains the application data gathered durring installation which is necesary to prepare employee data or process deductions
+    '''
+    auto_enroll: bool
+    group_id: str
+    
+
+class Name:
```
stevevls, line 16: I think this redefines aspects of the `Employee` common type.  Do we need it?

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
```diff
@@ -0,0 +1,109 @@
...
+class BenefitsEligibility(Enum):
+    # this will expand to cover cobra later
+    ELIGIBLE = 1
+    IN_ELIGIBLE = 2
+
+class BenefitsEligibilityStatus:
+    eligibility: BenefitsEligibility
+    effective_date: datetime
+
+class EmployeeStatus:
```
stevevls, line 88: Can this be sourced from `flux_core`?

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
```diff
@@ -0,0 +1,109 @@
...
+    status: EmployeeState
+    effective_date: datetime
+
+class EmployeeData:
```
stevevls, line 92: Source this from `flux_core`

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
```diff
@@ -0,0 +1,109 @@
...
+    hours_per_week: Optional[int]
+
+class Employment:
+    hours: EmploymentHours
+    is_rehire: bool
+    termination_date: Optional[datetime]
+    start_date: datetime
+    original_hire_date: datetime
+    w2_start_date: datetime
+    pay: Pay
```
stevevls, line 76: I imagine that much of this can be moved into flux core.  What are your thoughts @Kingpin007 ?

flux_sdk/ben_admin/capabilities/report_employee_data/interface.py
```diff
@@ -0,0 +1,25 @@
+from abc import ABC, abstractmethod
+
+from flux_sdk.flux_core.data_models import File
+from flux_sdk.ben_admin.capabilities.report_employee_data.data_models import AppConfig, EmployeeData
+
+
+class ReportEmployeeData(ABC):
```
stevevls, line 7: What sort of data is getting reported?  Let's be more specific in the name.

flux_sdk/ben_admin/capabilities/report_employee_data/interface.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    This class represents the "report employee data" capability. The developer is supposed to implement
+    the following methods in their implementation. 
+
+    The methods on this class are required to instantiate an instance of this class.
+    """
+
+    @staticmethod
+    @abstractmethod
+    def get_formated_file(app_config: AppConfig, employee_data: EmployeeData) -> File:
```
stevevls, line 17: Let's rename to `format_xxx_file` where `xxx` tells us what is actually in the file.

stevevls (commented):
Thanks for being one of our early contributors as we're still ironing out the process. 😅  This repo is public-facing, so this is going to be a very detailed review since 1) third party developers will need to be able to read it and make sense of it and 2) once merged, it's hard to change.  Thanks for bearing with us!

To start, I have a couple of high level comments:
1. Please draft a kit specification here: https://rippling.atlassian.net/wiki/spaces/ID/pages/3703538906/Kit+Specifications.  This spec should give us a clear breakdown of capabilities, what is provided by the kit, and what we require from a third party.
2. Can we rename this to be benefits admin to avoid jargon-y sounding names?
3. Left some comments inline, but in general please prefer existing data types from the `flux_core` package instead of creating your own.  Definitely open to chat if you there's a reason they're not suitable.
4. Also looks like in some cases you're defining some types that might want to get pushed down into `flux_core`.

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
bryanbierce: there is one field that is kit-level, it wasn't clear that there is an intended place for that. Is it `benifits_administration/data_models.py`?

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
bryanbierce: Not if we re-use `Employee` here.

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
bryanbierce: I agree, alot of this feels easily re-usable. The `Employement` object conflicts with `Employee` a bit as they have some overlappigng data.

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
bryanbierce: `EmployeeState` is, I was just wrapping it up with its related `effective_date` to make my `EmployeeData` tidier its just going to have to become `employee_state_effective_date` if I re-use `Employee`

flux_sdk/ben_admin/capabilities/process_deduction_data/interface.py
bryanbierce: Yeah I imagine that should be doable, the app developer can config that

bryanbierce:
> Thanks for being one of our early contributors as we're still ironing out the process. 😅 This repo is public-facing, so this is going to be a very detailed review since 1) third party developers will need to be able to read it and make sense of it and 2) once merged, it's hard to change. Thanks for bearing with us!
> 
> To start, I have a couple of high level comments:
> 
> 1. Please draft a kit specification here: https://rippling.atlassian.net/wiki/spaces/ID/pages/3703538906/Kit+Specifications.  This spec should give us a clear breakdown of capabilities, what is provided by the kit, and what we require from a third party.
> 2. Can we rename this to be benefits admin to avoid jargon-y sounding names?
> 3. Left some comments inline, but in general please prefer existing data types from the `flux_core` package instead of creating your own.  Definitely open to chat if you there's a reason they're not suitable.
> 4. Also looks like in some cases you're defining some types that might want to get pushed down into `flux_core`.
> 5. A lint check is failing...you can run `./run_ruff --fix` in the top level directory to auto-fix where possible.  It will also tell you where there are any lint violations that it wouldn't auto-fix.

Happy to help iron out the kinks :) 
1. yes, https://rippling.atlassian.net/wiki/spaces/ID/pages/3757047835/Benefits+Administration
2. yes
3. I'm not in love with the `Employee` object. Its does not contain all of the Role and Job data which we require which means I will need to extend it. Since it is so flat, the field names will end up being a bit messy. Looking in the flux kits code in `rippling-main` I don't see that there is any easily reusable constructors for `Employee` either so I'm just going to end up having to re-implement the builder anyways. I don't really see the point in reusing it. That said, I can re-implement that builder in a way that others can use it in the future and we can just deal with an un-tidy schema if that is a strong preference.
4. Yes, the data around pay and hours all feels like things that other kits might need in the future and should be maintained centrally if Employee is. 
5. I don't see that `./run-ruf` exists in `flux_sdk`. Is there a setup script that I missed? [Looking at the lint output](https://github.com/Rippling/flux-sdk/actions/runs/6868621016/job/18679586578?pr=18) I see that its mostly line length issues. Is there a config in `flux-sdk` that can be used to setup file formatting in VSCode or PyCharm?

flux_sdk/benefits_administration/capabilities/process_employees_deductions/interface.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    This class represents the "process employee deductions" capability. The developer is supposed to implement
+    the following methods in their implementation. 
+
+    The methods on this class are required to instantiate an instance of this class.
+    """
+
+    @staticmethod
+    @abstractmethod
+    def process_employees_deductions_into_rippling(deductions_file: File):
```
bryanbierce, line 16: @stevevls I am still waiting for some product clarity before I can finalize the type for the return on this one. To avoid churn on the type, are you open to me landing just the `ReportEmployeesPersonalAndEmploymentData` capability first and then adding this `ProcessEmployeeDeductions` in a later diff once I have clarity?

flux_sdk/benefits_administration/capabilities/report_employees_personal_and_employment_data/data_models.py
```diff
@@ -0,0 +1,109 @@
...
+    This contains the core data about an employee which is relevant to a ben admin provider
+    '''
+    id: str
+    ssn: str
+    name: Name
+    dob : datetime
+    phone_number: str
+    business_email: str
+    personal_email: str
+    gender: Gender
```
Kingpin007, line 103: change type to `Optional[Gender]`

flux_sdk/benefits_administration/capabilities/process_employees_deductions/interface.py
stevevls: Absolutely.  Feel free to yank this for now and we can add it in whenever you're ready.  Much better to wait and see in these instances. 😄

flux_sdk/ben_admin/capabilities/process_deduction_data/interface.py
stevevls: Eventually we want to be able to support a more turnkey setup where we provide parsers, de/compression, etc.  But we're not quite there yet.

The `get_file_name` was previously something we thought we would use to create a file matcher for downloading from SFTP.  But I think we have yet to find a case where this would need code.  So I would nix it unless we have some concrete use cases that couldn't be solved with configuration and regexes.  We don't want to unnecessarily complicate the interface to developers.  Less is more in these cases. 😄

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
stevevls: Ultimately that's where it should be.  But for now we have been dropping config classes into Rippling Main like [this one](https://github.com/Rippling/rippling-main/blob/master/app/apps/flux_runtime/config/kit_401k.py).  This is something we're going to fix on our end but not sure exactly yet of the timing so this is the pattern we'll go with for now.

flux_sdk/benefits_administration/capabilities/report_employees_personal_and_employment_data/interface.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    This class represents the "report employee data" capability. The developer is supposed to implement
+    the following methods in their implementation. 
+
+    The methods on this class are required to instantiate an instance of this class.
+    """
+
+    @staticmethod
+    @abstractmethod
+    def get_formated_employees_peronal_and_employment_data_file(config: ReportEmployeesPersonalAndEmploymentDataConfig, employee_data: list[EmployeePersonalAndEmploymentData]) -> File:
```
stevevls, line 17: Let's remove the `get_` and use `format_` as the verb.  I think it makes the intention of the function a little clearer.  I think we can eliminate the `_file` at the end too since that's implied by the return type.

```suggestion
    def format_employees_peronal_and_employment_data(config: ReportEmployeesPersonalAndEmploymentDataConfig, employee_data: list[EmployeePersonalAndEmploymentData]) -> File:
```

stevevls:
> I'm not in love with the Employee object. Its does not contain all of the Role and Job data which we require which means I will need to extend it. Since it is so flat, the field names will end up being a bit messy. Looking in the flux kits code in rippling-main I don't see that there is any easily reusable constructors for Employee either so I'm just going to end up having to re-implement the builder anyways. I don't really see the point in reusing it. That said, I can re-implement that builder in a way that others can use it in the future and we can just deal with an un-tidy schema if that is a strong preference.
This is a fair point.  If you think it makes sense to create a specialized set of classes for your kit, that's totally fine.  They would just need to be named differently than the core types so that the name indicates their purpose.  I feel like re-using the name `Employee` in different contexts would be confusing.

> I don't see that ./run-ruf exists in flux_sdk. Is there a setup script that I missed? [Looking at the lint output](https://github.com/Rippling/flux-sdk/actions/runs/6868621016/job/18679586578?pr=18) I see that its mostly line length issues. Is there a config in flux-sdk that can be used to setup file formatting in VSCode or PyCharm?
I think it was merged to `main` after you branch, so if you merge or rebase you should get it.  We don't have configs for IDEs (yet), but that's a good idea.

bryanbierce:
> This is a fair point. If you think it makes sense to create a specialized set of classes for your kit, that's totally fine. They would just need to be named differently than the core types so that the name indicates their purpose. I feel like re-using the name Employee in different contexts would be confusing.

I've gone ahead and updated the name on mine and used the model from core for part of it so we achieve some consistency. I am duplicating a few of the fields from the `Employee` object within `Employment` and `status` since I think it will be more natural for our consumers to read these types and find that data in those places

flux_sdk/benefits_administration/capabilities/report_employees_personal_and_employment_data/data_models.py
```diff
@@ -0,0 +1,90 @@
...
+
+class EmployeePersonalAndEmploymentData:
+    '''
+    This contains the core data about an employee which is relevant to a benefits administration provider
+    '''
+    personal: Employee
+    employment: Employment
+    pay: Pay
+    status: EmployeeStatus
+    benefits_eligibility: BenefitsEligibilityStatus
```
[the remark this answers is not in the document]
bryanbierce: That's a good question. Currently we have no use case for prior status'. I'm struggling to think of the business use case that would require it. 🤔 Although, it might be useful is in supporting error recovery. Say we have an outage of some sort and the data is inaccurate for a period, we might want to be able to send a historical correction. I can nest that status into a `Benefits` object and leave room for that possibility

stevevls:
@bryanbierce I just merged in latest main because we've added lint and test actions in GH that are now required statuses to merge.  Looks like we have some linting failures to take care of.  The mypy status check isn't required yet b/c some errors got merged down to main, but please make sure that we don't introduce any new issues as we plan to make it required ASAP.  Thanks!

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
```diff
@@ -0,0 +1,24 @@
...
+from typing import Optional
+
+
+class DeductionDetails:
+    companyId: str
+    employeeId: str
+    deductionCode: str
+    effectiveFrom: datetime
+    endDate: datetime
+    employeeContribution: Optional[Decimal]
```
Kingpin007, line 13: Replace with `Decimal | None`

flux_sdk/benefits_administration/capabilities/report_employees_personal_and_employment_data/data_models.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class Pay:
+    frequency: PayFrequency
+    frequency_effective_date: datetime
+    time_unit: PayTimeUnit
+    value_per_unit: MonetaryValue
+    value_effective_date: datetime
+    salary_or_equivalent: MonetaryValue
+    expected_commission: Optional[MonetaryValue]
```
Kingpin007, line 56: `MonetaryValue | None`

flux_sdk/benefits_administration/capabilities/report_employees_personal_and_employment_data/data_models.py
```diff
@@ -0,0 +1,111 @@
...
+
+class EmploymentHours:
+    type: EmploymentType
+    type_effective_date: datetime
+    hours_per_week: Optional[int]
```
Kingpin007, line 61: `int | None`

flux_sdk/benefits_administration/capabilities/report_employees_personal_and_employment_data/data_models.py
```diff
@@ -0,0 +1,111 @@
...
+    ABANDONMENT = 4
+    OFFER_DECLINE = 5
+    RESCIND = 6
+    RENEGE = 7
+
+class Employment:
+    hours: EmploymentHours
+    pay: Pay
+    is_rehire: bool
+    termination_date: Optional[datetime]
```
Kingpin007, line 78: `datetime | None`

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
bryanbierce: Sure. But, why? They're the same thing. An `Optional` is a normal and concise way of expressing this, common in modern languages. Why write the longhand union?

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
```diff
@@ -0,0 +1,23 @@
...
+from decimal import Decimal
+from enum import Enum
+
+
+class DeductionDetails:
+    companyId: str
+    employeeId: str
+    deductionCode: str
+    effectiveFrom: datetime
+    endDate: datetime
```
stevevls, line 11: Should this be ` | None`?

flux_sdk/benefits_administration/capabilities/process_employees_deductions/interface.py
```diff
@@ -0,0 +1,36 @@
...
+    the following methods in their implementation. 
+
+    The methods on this class are required to instantiate an instance of this class.
+    """
+
+    @staticmethod
+    @abstractmethod
+    def get_file_name(company_name: str) -> str:
+        """
+        This method returns the file name of deductions sent by Bswfit
```
stevevls, line 22: IIUC, Bswift is a particular vendor and doesn't represent the general case.  Is there a reason we need this as a method instead of as configuration (e.g. with a filename mask) that could be used against the remote SFTP server?

stevevls (commented):
Looking back at my comments from last fall, it still seems like we have a lot of potential for duplication of data types with those in flux core.  Would be curious to have @pshanmukrao or @Kingpin007's opinion on whether we should be factoring some of them into the shared location.

flux_sdk/benefits_administration/capabilities/process_employees_deductions/interface.py
Kingpin007: +1

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
```diff
@@ -0,0 +1,111 @@
...
+    CONTRACTOR = 1
+    SALARIED_FT = 2
+    SALARIED_PT = 3
+    HOURLY_FT = 4
+    HOURLY_PT = 5
+    TEMP = 6
+
+
+class MonetaryValue:
+    value: float
```
Kingpin007, line 27: should this be Decimal instead?

flux_sdk/ben_admin/capabilities/report_employee_data/data_models.py
Kingpin007: If there is additional information in `Employment` I would recommend you extend `Employee` and add it to flux-core

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
Kingpin007: https://peps.python.org/pep-0604/
I read in some discussion forums for PEP that the `Optional` syntax would be deprecated in a future python release in a few years.

flux_sdk/benefits_administration/capabilities/process_employees_deductions/interface.py
bryanbierce: Well that depends on who/where we would be setting up the file name. Usually the file name is going to be built using the company data in some way, such as the `GroupID` which we collect from the admin during app install, or the company id. What does the support for configuring the filename look like right now? If today, Bswift wants to build the filenames from a pattern they use that includes the `GroupId` and tomorrow the next 3rd party wants to use the company id how would I set that up through configuration today?

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
bryanbierce: Ohhh interesting, I had no idea. Thanks for sharing :) That seems a shame to me but 🤷 not my language. And hopefully we won't be using it by then anyways 🤞

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
```diff
@@ -0,0 +1,112 @@
...
+    '''
+    This contains the application data gathered durring installation
+    which is necesary to prepare employee data or process deductions
+    '''
+    auto_enroll: bool
+    group_number: str
+    company_id: str
+    
+
+class EmploymentType(Enum):
```
Kingpin007, line 18: Move this to `submodules/flux-sdk/flux_sdk/flux_core/data_models.py` as this can be reused

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
```diff
@@ -0,0 +1,112 @@
...
+    HOURLY_PT = 5
+    TEMP = 6
+
+
+class MonetaryValue:
+    value: Decimal
+    currency_type: str
+
+
+class PayFrequency(Enum):
```
Kingpin007, line 32: Move to `submodules/flux-sdk/flux_sdk/flux_core/data_models.py` as this can be reused

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
```diff
@@ -0,0 +1,112 @@
...
+    WEEKLY = 1
+    BI_WEEKLY = 2
+    MONTHLY = 3
+    SEMI_MONTHLY = 4
+    QUARTERLY = 5
+    ANNUALLY = 6
+
+class PayTimeUnit(Enum):
```
Kingpin007, line 40: Same move to `submodules/flux-sdk/flux_sdk/flux_core/data_models.py`

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
```diff
@@ -0,0 +1,112 @@
...
+    salary_or_equivalent: MonetaryValue
+    expected_commission: MonetaryValue | None
+
+class EmploymentHours:
+    type: EmploymentType
+    type_effective_date: datetime
+    hours_per_week: int | None
+    hours_effective_date: datetime
+
+class TerminationType(Enum):
```
Kingpin007, line 64: Same move to `submodules/flux-sdk/flux_sdk/flux_core/data_models.py`

Kingpin007 (approved):

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
```diff
@@ -0,0 +1,23 @@
+from datetime import datetime
+from decimal import Decimal
+from enum import Enum
+
+
+class DeductionDetails:
+    companyId: str
```
sz2376, line 7: We have been using `company_id` snake case in this repo rather than `companyId` camel case. You are also mixing these two formats in this PR thus I'd like to request a change for these.

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/interface.py
```diff
@@ -0,0 +1,31 @@
...
+            config: ReportEmployeesHrDataConfig,
+            employee_data: list[EmployeeHrData]
+    ) -> File:
+        """
+        This method receives the apps configuration data and a list of employee data.
+        The developer is expected to create a file, formated to their use case, and return that file.
+        The file will be tranfered to the partner company via SFTP
+
+        :param App
+        :param list[EmployeeData]:
```
sz2376, line 29: The docstring does not match with the actual parameters

flux_sdk/benefits_administration/capabilities/process_employees_deductions/interface.py
```diff
@@ -0,0 +1,27 @@
...
+    @staticmethod
+    @abstractmethod
+    def format_and_fetch_deduction_info(
+            uri: str, stream: IOBase, deduction_code_mapping: list[ExternalDeductionCodeToRipplingCode]
+    ) -> list[DeductionDetails]:
+        """
+        This method receives the file which contains the deductions relevant to the companies employees and returns the
+        deductions details for each employee
+        :param File:
+        :return:
```
sz2376, line 26: This does not look good. These docstrings are important for us to generate the documentation so please get them as accurate as possible.

sz2376 (dismissed):

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
```diff
@@ -0,0 +1,84 @@
...
+
+class Employment:
+    hours: EmploymentHours
+    pay: Pay
+    is_rehire: bool
+    termination_date: datetime | None
+    termination_type: TerminationType | None
+    start_date: datetime
+    original_hire_date: datetime
+    w2_start_date: datetime
```
pshanmukrao, line 54: @bryanbierce What are the fields do you need that are not part of Employee ? 

I see that `hours_per_week` and `hours_effective_date` are the only fields ?  Can they go inside Employee data type ? cc: @stevevls @Kingpin007 @ujjwalshukla

sz2376 (approved):

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
bryanbierce: `type_effective_date`, `hours_per_week`, `hours_effective_date` , `termination_type` and the `pay` object are not on `Employee` currently. I had asked the team if I could extend `Employee` with these originally and the answer from @Kingpin007 and @ujjwalshukla was no. Happy to move them around if you all would rather co-locate these now. This `Employment` object only existed because the team originally did not want any of the new data in `flux_core` so I took the opportunity to duplicate some fields and make the objects a bit more sensible to work with.

PR #23, opened 2023-12-06
WIP: add etl kit

dominicbarnes:
This adds the first iteration of our ETL (aka: "Data Import") kit, which includes 1 capability for "Single Object Import". Most of the work here was carrying over the details from our internal planning and translating them into data models and the interface.

Since this kit is a new endeavor rather than adapting patterns we've used in previous apps, this should be regarded as experimental and subject to breaking changes until we've actually deployed a working implementation.

**NOTE:** this PR will be left unmerged while we develop our first implementation.


flux_sdk/data_import/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Use this hook to construct the query that should be run by Rippling to extract records.
+        :param schema: The schema generated in the "get_schema" hook for this object.
+        :param checkpoint: If included, this is an incremental sync and should adjust the Query accordingly.
+        :return: Query
+        """
+
+    @staticmethod
+    @abstractmethod
+    def process_records(schema: Schema, records: list[Record]) -> list[Record]:
```
sz2376, line 37: Shall we have a way for the app developer to indicate that no post processing needed? Otherwise they would implement this hook as:

```
def process_records(schema: Schema, records: list[Record]) -> list[Record]:
  return records
```

flux_sdk/data_import/data_models/query.py
```diff
@@ -0,0 +1,34 @@
...
+    # This is always required and indicates which connector type is being used.
+    type: QueryType
+
+    # This is required for SQL connectors. It is the raw text of the query to be executed.
+    sql_query: str
+
+    # This is optional for SQL connectors. It is used to add safe interpolation of values into the query, rather than
+    # requiring it to be constructed internally. In the sql_query, each "@var" will be replaced with the corresponding
+    # "var" from this dict.
+    sql_query_args: dict[str, Any]
```
sz2376, line 28: Should the value be of str type?

flux_sdk/data_import/data_models/query.py
```diff
@@ -0,0 +1,34 @@
...
+
+    # This is required for MONGODB connectors. It indicates which collection the query will be executed in.
+    mongo_collection: str
```
sz2376, line 31: Is the database configured as part of the data connector? @stevevls Otherwise we would also need the database in the hook response.

stevevls (commented):
Good start.  I believe the plan is to name this ETL kit, though.  Let's check with Sam!

flux_sdk/data_import/data_models/record.py
```diff
@@ -0,0 +1,35 @@
+from dataclasses import dataclass
+from datetime import date, time, datetime
+from typing import Union, Optional
+
+
+# This represents the currently supported types for Record fields. Currently, we only support these primitive types and
+# do not allow for any complex/nested types.
+SupportedTypes = Union[str, int, float, bool, date, time, datetime, None]
```
sz2376, line 8: Is there any specific reason for the None type supported?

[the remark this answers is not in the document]
dominicbarnes: This allows an explicit `NULL` value to be passed along, which will be represented in fields as `None`.

I'm open to the rename, I'll apply that momentarily.

flux_sdk/data_import/data_models/record.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+# This corresponds to a row in the source database.
+@dataclass
+class Record:
+    # The ID for this record, which becomes external_id in the Custom Object.
+    primary_key: str
+
+    # This is the rest of the raw data for this field.
```
sz2376, line 17: field -> record

flux_sdk/data_import/data_models/query.py
```diff
@@ -0,0 +1,34 @@
...
+
+    # This is optional for MONGODB connectors. It provides the filter criteria for a basic query.
+    mongo_filter: dict[str, Any]
```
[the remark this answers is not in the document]
dominicbarnes: Yeah, that is one use-case. It could also just be used to just filter the dataset for other reasons specific to their use-case.

flux_sdk/data_import/data_models/record.py
```diff
@@ -0,0 +1,35 @@
...
+# This represents the currently supported types for Record fields. Currently, we only support these primitive types and
+# do not allow for any complex/nested types.
+SupportedTypes = Union[str, int, float, bool, date, time, datetime, None]
+
+
+# This corresponds to a row in the source database.
+@dataclass
+class Record:
+    # The ID for this record, which becomes external_id in the Custom Object.
+    primary_key: str
```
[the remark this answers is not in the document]
dominicbarnes: I don't believe CO supports that, for numeric types they'll be converted to strings anyway.

flux_sdk/data_import/data_models/record.py
```diff
@@ -0,0 +1,35 @@
...
+
+    # This is the rest of the raw data for this field.
+    fields: dict[str, SupportedTypes]
+
+    # This represents the links in this object, which correspond to the Schema references.
+    references: list[Optional[str]]
```
sz2376, line 21: I was not too sure how the references field is used here. Is there any example for the references field?

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,92 @@
...
+    Currency = 9
+    # Embedded = 10
+    # MultiEnum = 11
+    Percent = 12
+    LongText = 13
+    # Index = 14
+    # PolymorphicEdge = 15
+    Email = 16
+    Url = 17
+    Time = 18
```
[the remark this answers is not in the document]
dominicbarnes: Yeah, these correspond to this enum: https://github.com/Rippling/rippling-main/blob/e07c196990093046fff160c04ddc839a4dab012f/app/hub_platform/universal_objects/fields/types.py#L77-L97

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # This is the name of the field on the linked object to use.
+    lookup: str
+
+
+# A schema describes the shape of an object being imported and is used by Rippling to define the Custom Object.
+@dataclass
+class Schema:
+    # This is the name of the object.
+    name: str
```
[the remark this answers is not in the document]
dominicbarnes: Yeah, this is the name of the Custom Object, I can make that more clear.

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,92 @@
...
+@dataclass
+class SchemaReference:
+    # This is the name of the field on this object that is being used for the link.
+    origin: str
+
+    # This is the name of the object being linked to.
+    object: str
+
+    # This is the name of the field on the linked object to use.
+    lookup: str
```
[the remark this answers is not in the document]
dominicbarnes: I anticipate we'll validate the returned value before even opening a connection so I _think_ this problem can be alleviated. That being said, I'm open to the classes approach too, let me tinker with it.

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,92 @@
...
+    references: list[SchemaReference]
+
+    # These are the remaining data fields.
+    fields: list[SchemaField]
+
+
+# These are the available lookup-fields for an employee lookup.
+class EmployeeLookup(StrEnum):
+    EMPLOYEE_ID = "employee_id"
+    BUSINESS_EMAIL = "business_email"
```
sz2376, line 81: Just a name convention that we usually call it work email

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,92 @@
...
+    PERSONAL_EMAIL = "personal_email"
+
+
+# This is a helper for creating a link/reference to a Rippling employee.
+@dataclass
+class EmployeeRef(SchemaReference):
```
sz2376, line 87: How do people specify the origin field? Should there be a constructor which takes in the origin field?

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,92 @@
...
+    # This uses an enum as a helper.
+    lookup: EmployeeLookup
+
+    # This sets a default value to remove the need to define explicitly.
+    object: str = "employee"
```
sz2376, line 92: The API name for the employee module is `employee__c` if we want to follow the API name rather than just the name.

flux_sdk/data_import/data_models/record.py
```diff
@@ -0,0 +1,35 @@
...
+
+    # This flag can be used by the "process_records" hook to signal to Rippling that the object should not be imported.
+    # If a Record is not found after this hook, that will be regarded as an error, so this flag should be used instead.
+    drop: bool
+
+    # This is used by the "process_records" hook to signal to Rippling the checkpoint that can be observed and recorded
+    # after the sync is complete.
+    #
+    # In the case of datetime/int values, Rippling will keep track of the *highest* value seen from all records. In the
+    # case of str values, Rippling will only record the *last* value seen.
```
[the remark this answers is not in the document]
dominicbarnes: The use-case described by @sz2376 that we use for MongoDB seems to assume that the underlying `ObjectId` is unique but not used for sorting, so I'm not sure how to work around this for the use-case we're probably building for in the first version.

flux_sdk/data_import/data_models/record.py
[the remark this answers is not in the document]
stevevls: Ah...got it.  So everything is essentially nullable?

flux_sdk/data_import/data_models/query.py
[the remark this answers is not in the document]
stevevls: Okay.  Let's add some docs then that this is where the checkpoint would be incorporated.

flux_sdk/data_import/capabilities/single_object_import/interface.py
dominicbarnes: Yeah, that was an open question I had, is there a way to mark hooks as optional? I wonder if this would need to be pulled out into it's own capability.

flux_sdk/data_import/data_models/record.py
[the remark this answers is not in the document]
stevevls: Okay.  So then do we need rules around type co-ercion in case the PK field in the result set is not a string?

flux_sdk/data_import/data_models/query.py
dominicbarnes: We should probably have a limited set of values here too, but I can easily see `datetime | int | str` as a place to push the checkpoint to.

flux_sdk/data_import/data_models/record.py
dominicbarnes: This allows for an explicit `NULL` to be handled gracefully, which I imagine we need to support.

flux_sdk/data_import/data_models/record.py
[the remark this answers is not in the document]
stevevls: Sounds good...positional seems problematic to me but also understand the desire to keep it clean.  Any easy way to tie it back to the declared schema?

flux_sdk/data_import/data_models/schema.py
dominicbarnes: I noticed in `flux_core` there is an `Employee.business_email`, so I was following that convention here too.

flux_sdk/data_import/data_models/schema.py
dominicbarnes: `@dataclass` allows you to specify any of the underlying attributes as kwargs in an automatically-generated `__init__` so there's no need to explicitly define it.

flux_sdk/data_import/data_models/record.py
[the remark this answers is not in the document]
stevevls: Hmmm...I don't know if that means we should relax the ordering requirement for any type of string, though.  If anything, I think it would be the Mongo app author's responsibility to make a compound sortable cursor for us to use.  Otherwise, I think we start to make compromises around parallel processing of records in order to guarantee the "last" value.  Plus I think it's just confusing in general to have a non comparable cursor. 😅

flux_sdk/data_import/data_models/schema.py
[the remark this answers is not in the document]
stevevls: Got it.  Still unsure whether we'd want to have CO details leaked out in the SDK.  I think I'd opt for strings still and then translate to enum values in the kit.

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+# A schema describes the shape of an object being imported and is used by Rippling to define the Custom Object.
+@dataclass
+class Schema:
+    # This is the name of the object.
+    name: str
+
+    # This indicates which field (by name) is used as the ID for the object.
+    primary_key: str
```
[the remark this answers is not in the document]
stevevls: Okay.  Should we leave a comment then that only string keys are supported?  And any co-ercion rules that may come into play?

flux_sdk/data_import/data_models/schema.py
sz2376: Cool then let's stick to the `business_email` if that's already there in the Employee class. Hopefully people don't introduce the `work_email` later.

flux_sdk/data_import/capabilities/single_object_import/interface.py
stevevls: I think this is where we would introduce the concept of optional hooks.  I don't think we'd want to invoke the Lambda if it's just going to echo back the input. 😅

flux_sdk/data_import/data_models/query.py
stevevls: @sz2376 by database you mean the "catalog" and not the actual connection, right?  I think it would make sense for that to be configured on the connection so no need to put it in the hook.

flux_sdk/data_import/data_models/schema.py
stevevls: I don't think there's an `empoyee__c` type, is there?  the `__c` indicates a custom object but the employee type is meant to be RWC, no?

flux_sdk/data_import/data_models/record.py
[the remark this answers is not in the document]
dominicbarnes: That's my thinking. That being said, there is a way to mark a field as "required" in CO, which I think implies that some fields can be null/empty by default, but I can double-check that assumption.

flux_sdk/data_import/capabilities/single_object_import/interface.py
sz2376: I don't think we have an existing mechanism to detect whether the hook is implemented or not for the app even though it's technically doable. One approach I can think of is to let the kit open up a configuration variable to let the app developer specify. An example here: https://github.com/Rippling/flux-apps/blob/main/flux_apps/principal_v2/manifest.json#L143

flux_sdk/data_import/data_models/schema.py
sz2376: Yeah a nitpicking is that people might forget to specify the origin field which produces a functionally incorrect object but the build still passes. Maybe a predefined `__init__` function makes sure people specify the origin field?

flux_sdk/data_import/data_models/schema.py
sz2376: The `employee__c` is essentially the api_name for the custom object representing the Employee module so we can use `ee_co = get_custom_object(company=company, api_name=EE_CO_API_NAME)` to fetch the CO for the Employee module. The api_name is a unique field on the CustomObject record so I was assuming we want to reference by the api_name rather than the name, which is not unique. Open to any clarifications if I misunderstood this.

stevevls: I kind of feel like we're talking past each other.  I don't think of employee as a custom object.  When I think about resolving to an employee, I'm imagining that under the hood we're mapping to RWC.  As far as I know, there's no `employee__c` custom object that would be available in every account because...well...it wouldn't be a custom object if everyone had it.  But maybe I'm misunderstanding?

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,79 @@
...
+    is_required: bool
+
+    # When enabled, this value will be enforced as unique by Rippling Custom Object.
+    is_unique: bool
+
+
+# These are the available lookup-fields for an employee lookup.
+class EmployeeLookup(StrEnum):
+    EMPLOYEE_ID = "employee_id"
+    BUSINESS_EMAIL = "work_email"
```
sz2376, line 44: I somehow still feel we should keep the name as WORK_EMAIL so it keeps consistency here at least. We should change the business_email to work_email in the [Employee](https://github.com/Rippling/flux-sdk/blame/main/flux_sdk/flux_core/data_models.py#L83) class before it's too late.

sz2376 (commented):
I think we want to change the data_import folder name to etl

flux_sdk/data_import/data_models/schema.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+# This lists the available types that can be used for schema references.
+ReferenceType = Union[CustomObjectReference, EmployeeReference]
+
+
+# A schema describes the shape of an object being imported and is used by Rippling to define the Custom Object.
+@dataclass
+class Schema:
+    # This is the name of the Custom Object.
```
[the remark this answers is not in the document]
dominicbarnes: Not yet, but we can just add another field for that.

flux_sdk/data_import/data_models/schema.py
dominicbarnes: I've raised this question internally in Slack to see how we should proceed.

flux_sdk/etl/data_models/query.py
```diff
@@ -0,0 +1,33 @@
...
+class MongoQuery:
+    # This indicates which collection the query will be executed in.
+    collection: str
+
+    # This provides the filter criteria for a basic query.
+    filter: dict[str, Any]
+
+
+# This is the list of types that can be used to represent a query.
+QueryType = Union[SQLQuery, MongoQuery]
```
stevevls, line 33: Is the `Type` at the end a python convention?  For some reason, adding that at the end makes me expect this to be an enum, so I'd suggest we drop the `Type` unless that would make the code not idiomatic.

flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,45 @@
...
+        :return: Schema
+        """
+
+    @staticmethod
+    @abstractmethod
+    def prepare_query(schema: Schema, checkpoint: Optional[CheckpointType]) -> QueryType:
+        """
+        Use this hook to construct the query that should be run by Rippling to extract records.
+        :param schema: The schema generated in the "get_schema" hook for this object.
+        :param checkpoint: If included, this is an incremental sync and should adjust the Query accordingly.
```
stevevls, line 31: Is the assumption here that if an app doesn't provide checkpoints that it's not possible to run in incremental mode?

flux_sdk/etl/data_models/query.py
stevevls: Rather than drop a comment on all the *Type types, consider this a blanket statement. :)

flux_sdk/etl/data_models/query.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+# This is returned by the "prepare_query" hook for SQL connectors.
+@dataclass
+class SQLQuery:
+    # This is the raw text of the query to be executed.
+    text: str
+
+    # This is used to add safe interpolation of values into the query, rather than requiring it to be constructed by the
+    # hook. In the sql_query, each "@var" will be replaced with the corresponding "var" from this dict.
```
stevevls, line 16: Do we need to define an escaping scheme for the `@`?

flux_sdk/etl/data_models/record.py
```diff
@@ -0,0 +1,39 @@
...
+
+# This indicates what types can be used for a primary key. Using a string is recommended, as numeric values will be
+# coerced into strings within Rippling.
+PrimaryKeyType = Union[str, int]
+
+# This represents the currently supported types for Record fields. Currently, we only support these primitive types and
+# do not allow for any complex/nested types.
+FieldType = Union[str, int, float, bool, date, time, datetime, None]
+
+# Checkpoints can be represented as either timestamp (datetime), sequence number (int) or an opaque token (str).
```
stevevls, line 13: Let's call out that this requires a strict LT/GT semantic in order to function correctly.

flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    This class represents the "single_object_import" capability. This is the first iteration of our Data Import
+    framework, but is limited in scope to only operating on a single object at a time. This is a stepping stone towards
+    a more robust system, so consider this experimental.
+    """
+
+    @staticmethod
+    @abstractmethod
+    def get_schema() -> Schema:
+        """
```
stevevls, line 19: I'm guessing now isn't an appropriate time to talk about _implementing_ schema evolution, but should we drop a sentence or two explaining what would happen with compatible/incompatible changes?

flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @staticmethod
+    @abstractmethod
+    def process_records(schema: Schema, records: list[Record]) -> list[Record]:
+        """
+        Use this hook to post-process the Records extracted via the query performed from "prepare_query". This hook is
+        required in order to implement incremental sync (via Record.checkpoint) or to dynamically drop records from the
+        sync (via Record.drop).
+        :param schema: The schema generated in the "get_schema" hook for this object.
```
stevevls, line 42: Should we drop a sentence on expectations of the returned record in terms of needing to provide a valid record that conforms to the schema?  Do we need to say here what would happen if the record was invalid?

flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,45 @@
...
+    def get_schema() -> Schema:
+        """
+        Use this hook to indicate the structure (aka: schema) for the data being imported. This is used to ensure that
+        Rippling Custom Objects has a target for the extracted records to be loaded into.
+        :return: Schema
+        """
+
+    @staticmethod
+    @abstractmethod
+    def prepare_query(schema: Schema, checkpoint: Optional[CheckpointType]) -> QueryType:
```
stevevls, line 27: I'm still a little unconvinced on this method.  Since there is no connector metadata getting passed in, I could return e.g. SQL for a Mongo connection or Mongo for a SQL connection.  Some ideas (neither of which are perfect):
1. Pass in some connector metadata (connector type and maybe subtype, e.g. Mongo, SQL/Postgres)
2. Pass in a blank instance of `QueryType` and have the caller do instanceof checks to figure out what it needs to populate
3. Have different hooks altogether for the mongo and SQL cases.  And then the kit can decide which to invoke based on what connector it has.

Not completely sure about any of those, but maybe some of these ideas inspire you?

stevevls (commented):
Looks great.  Definitely onto minor comments at this point, though a few could still impact the API.

flux_sdk/etl/data_models/query.py
dominicbarnes: Honestly, I'm not sure what the convention is, let me search some of our codebases.

flux_sdk/etl/capabilities/single_object_import/interface.py
dominicbarnes: Yeah, that is my intention.

stevevls: Sounds good.  Let's call it out then.

flux_sdk/etl/data_models/query.py
dominicbarnes: The vast majority of definitions involving `Union[...]` have no such suffix or conventions that I can detect. I found a handful with _some_ sort of differentiation, such as `CONSTANT_CASE_TYPES` or `Types` suffix or even `T`, but there appears to be no consistency.

I'll drop the `Types` suffix for brevity.

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,96 @@
...
+    # This is a longer explanation of the field that will appear in the Rippling Custom Object UI.
+    description: str
+
+    # This is used to put this Custom Object into a group/category. If a category with this name already exists, a new
+    # one will not be created.
+    category_name: str
+
+    # When creating a new group/category, this will set a description that will appear in the Rippling Custom Object UI.
+    # If the category already exists, the description will not be updated.
+    category_description: str
```
sz2376, line 38: Ah I think these two need to be on the Schema class. 
> If the category already exists, the description will not be updated.

I guess we will update the description?

sz2376 (commented):
For all the optional fields w'd define them with `| None` as for example: `description: str | None`

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,96 @@
...
+    lookup: EmployeeLookup
+
+
+# This lists the available types that can be used for schema references.
+Reference = Union[CustomObjectReference, EmployeeReference]
+
+
+# A schema describes the shape of an object being imported and is used by Rippling to define the Custom Object.
+@dataclass
+class Schema:
```
sz2376, line 83: Customer can also specify the description for the custom object so we probably want a description field in the Schema as well.

flux_sdk/etl/data_models/record.py
```diff
@@ -0,0 +1,43 @@
...
+# This corresponds to a row in the source database.
+@dataclass
+class Record:
+    # The ID for this record, which becomes external_id in the Custom Object. Since only a simple, string value is
+    # supported by Custom Objects, this does not need to be a full SchemaField.
+    primary_key: PrimaryKey
+
+    # This represents the links in this object. Each key is the field name, which is used to look up the target details.
+    # Each value is the identifier used for the link.
+    references: dict[str, str]
```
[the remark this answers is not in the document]
dominicbarnes: Here's an example. Imagine this schema:

```python
references: {
  "other_object_id": CustomObjectReference(object="other_object", lookup="id")
}
```

Since the schema already specifies most of the edge/link information, the record is simplified:

```python
references: {
  "other_object_id": "object_1"
}
```

We can use the `other_object_id` key in `Record` to find the corresponding metadata in `Schema`, and then resolve that object.

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # When creating a new group/category, this will set a description that will appear in the Rippling Custom Object UI.
+    # If the category already exists, the description will be updated.
+    category_description: Optional[str]
+
+    # This indicates which field (by name) is used as the ID for the object.
+    primary_key: str
+
+    # These are the links to other objects. The keys are the field names that should be the origin for the link/edge.
+    references: dict[str, Reference]
```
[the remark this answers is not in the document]
dominicbarnes: Done!

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,129 @@
...
+class Schema:
+    # This is the name of the Custom Object.
+    name: str
+
+    # This is used to put this Custom Object into a group/category. If a category with this name already exists, a new
+    # one will not be created.
+    category_name: str
+
+    # This is the name of the field used for the primary key (aka: External ID).
+    primary_key_field: str
```
sz2376, line 102: The external id is not a required field so I think we can mark this as optional.

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # When enabled, this value will be enforced as unique by Rippling Custom Object.
+    is_unique: bool = False
+
+    # When using an Enum data type, this indicates the possible values.
+    enum_values: Optional[list[str]] = None
+
+    # When using an Enum data type, this indicates if *only* enum_values are accepted, meaning anything else will be
+    # rejected with an error.
+    enum_restricted: bool = False
```
sz2376, line 49: Should this also be optional as it only applies to enum types?

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,129 @@
...
+    description: Optional[str] = None
+
+
+# These are the available lookup-fields for a Rippling Employee.
+class EmployeeLookup(Enum):
+    # The Rippling Employee ID.
+    EMPLOYEE_ID = "employee_id"
+
+    # The work/business email for a Rippling Employee.
+    BUSINESS_EMAIL = "business_email"
```
sz2376, line 71: I guess our conclusion from this thread is to use `work_email` so that we keep it consistent with the platform API and the Rippling UI? https://rippling.slack.com/archives/C0589RX5QMA/p1701977151232189

flux_sdk/etl/data_models/schema.py
dominicbarnes: It effectively is optional since it includes a default value.

flux_sdk/etl/data_models/schema.py
dominicbarnes: I'm honestly not sure where we stand, nobody else weighed in. 🤷

flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,61 @@
...
+        required in order to implement incremental sync (via checkpoint) or to dynamically drop records from the sync
+        (via drop).
+
+        If the Record does not adhere to the provided schema, such as using an incorrect/incompatible type for a field,
+        the sync will fail. As such, this hook is the opportunity to massage any raw query results that may not be
+        directly compatible.
+
+        :param schema: The schema generated in the "get_schema" hook for this object.
+        :param records: This batch of records should be updated
+        :return: Query
```
falcon2212, line 60: nit: return type needs to be corrected

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,231 @@
...
+
+    # This is the name of the field used for the record name, which is displayed in the Rippling Custom Object UI. When
+    # not provided,
+    name_field: str
+
+    # These are the remaining data fields.
+    fields: list[SchemaField]
+
+    # This is the name of the field used for the primary key (aka: External ID).
+    primary_key_field: Optional[str] = None
```
falcon2212, line 171: Curious why is this field is optional? Also since we are using rippling custom objects shouldn't external_id will always be the name of the primary key?

dominicbarnes:
I'm closing this PR and will re-open one with a fresh comment/review history.

PR #28, opened 2023-12-13
[INSURANCE-67360] Added utility for vendors using Ascensus format

codewizzy:
JIRA: https://rippling.atlassian.net/browse/INSURANCE-67360
Added utility for vendors using Ascensus format.
Added unit test.


flux_sdk/pension/test_utils/test_pay_konnect/test_pay_konnect.py
pshanmukrao, line 1: I have a highlevel folder structure comment.

tests should be in the similar path to implementation code. 

for e.g., 
flux_sdk/pension/utils/ascensus.py 

should have tests in flux_sdk/pension/utils/tests/test_ascensus.py having this consistency will make it easier for checking where the tests for a given file.

flux_sdk/pension/utils/pay_konnect.py
```diff
@@ -268,8 +268,8 @@ def format_contributions_for_pay_konnect_vendor(
         :return: File
         """
         customer_partner_settings = payroll_upload_settings.customer_partner_settings
-        plan_id: str = customer_partner_settings.get("plan_id")
```
pshanmukrao, line 271: what is the reason to remove the type  ?

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,268 @@
...
+
+    @staticmethod
+    def get_file_name(payroll_upload_settings: PayrollUploadSettings) -> str:
+        """
+        Receives a PayrollUploadSettings from which the developer
+        is expected to return file name based on payroll_upload_settings
+        :param payroll_upload_settings:
+        :return: str
+        """
+        environment = payroll_upload_settings.customer_partner_settings.get("env", "TS")
```
pshanmukrao, line 106: What does TS mean here ?

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,268 @@
...
+        client_id = str(payroll_upload_settings.customer_partner_settings["client_id"])
+        payroll_run_id = payroll_upload_settings.payrun_info.payroll_run_id
+
+        if not environment or not client_id:
+            raise RuntimeError(
+                "Environment and client_id must be present to upload Vanguard file"
+            )
+        timestamp = datetime.datetime.now()
+        format_timestamp = timestamp.strftime("%m%d%Y.%H%M%S")
```
pshanmukrao, line 115: Do you need this to be formatted in a specific timezone ?

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,268 @@
...
+                        )
+                    else:
+                        current_date_of_term = ""
+                    prior_hire_date = employee.original_hire_date.strftime(
+                        STANDARD_DATE_FORMAT
+                    )
+                    prior_term_date = ""
+                    if (
+                        hasattr(employee, "termination_date")
+                        and employee.termination_date
```
pshanmukrao, line 179: Is this not the same as current_date_of_term  ?

flux_sdk/pension/test_utils/test_pay_konnect/test_pay_konnect.py
```diff
@@ -123,7 +120,7 @@ def set_nested_attribute(self, obj, attribute, value):
                 setattr(curr_obj, attribute, value)
             curr_obj = getattr(curr_obj, attribute, None)
 
-    def test_format_contributions_for_payKonnect_vendor_failure(self):
+    def test_format_contributions_for_payKonnect_vendor_failure(self) -> None:
```
pshanmukrao, line 123: I don't see a point for return type for test methods

flux_sdk/pension/test_utils/test_ascensus/test_ascensus.py
```diff
@@ -0,0 +1,168 @@
...
+            )
+
+    def test_get_file_name(self) -> None:
+        file_name = ReportPayrollContributionsAscensusUtil.get_file_name(
+            self.payroll_upload_settings
+        )
+        timestamp = datetime.now()
+        format_timestamp = timestamp.strftime("%m%d%Y.%H%M%S")
+        test_file_name = "{}_{}_{}_{}.csv".format(
+            "TS", "HISS00","54321", format_timestamp,
```
pshanmukrao, line 166: This is assuming that timestamp doesn't change between get_file_name() and next statement right ? Anyway, we can mock the time in testing ?

flux_sdk/pension/test_utils/test_pay_konnect/test_pay_konnect.py
codewizzy: This is required as it was failing mypy checks.

flux_sdk/pension/test_utils/test_ascensus/test_ascensus.py
codewizzy: Good point, will fix this.

flux_sdk/pension/utils/pay_konnect.py
codewizzy: This was failing on typecheck, modified it to a better solution.

flux_sdk/pension/utils/ascensus.py
codewizzy: For test we use "TS" enviroment.

flux_sdk/pension/utils/ascensus.py
codewizzy: Ascensus uses exact format as Vanguard which already exist in our system, in `payroll._401kproviders.vanguard.api.VanguardApi.get_filename` method there is no specific timezone is used for formatting so I think we should be good.

flux_sdk/pension/utils/ascensus.py
codewizzy: Here is the definition for prior_term_date and current_date_of_term as per ascensus
prior_term_date: If the employee was terminated and re-hired, list their prior termination date.
current_date_of_term: If the employee was terminated and re-hired and they re still employed leave this field blank. If the employee was terminated, re-hired, and terminated again, list the most recent termination date.

So this will be same if employee is currently terminated, if employee is non terminated then current_date_of_term is "".

vguptarippling: @codewizzy This terminate date doesn't return the old_terminated_date for rehires. You might need to create a new field for this. Please check actual data on prod and confirm

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,268 @@
...
+                    hours_worked = getattr(
+                        employee_payroll_record, "hours_worked", Decimal(0)
+                    )
+
+                    mapping_from_column_name_to_value = {
+                        "SOCIAL SECURITY": ssn,
+                        "LAST NAME": employee_last_name,
+                        "FIRST NAME": employee_first_name,
+                        "MI": "",
+                        "DIVISIONAL CODE": "",
```
vguptarippling, line 219: This will be a variable for the app. This needs to entered by customer during installation which is currently happening on our Vanguard app.

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,268 @@
...
+                        "TOTAL COMPENSATION": gross_pay,
+                        "EMPLOYEE 401(K)": payroll_employee_contribution_401k,
+                        "ROTH 401(K)": payroll_employee_roth_401k,
+                        "LOAN PAYMENT AMOUNT": payroll_employee_loan_repayment,
+                        "MATCH": payroll_company_match_contribution,
```
vguptarippling, line 224: There will be another variable named as company_contribution_column. You are directly adding it to match column. This should be configurable. Some companies might want it to SAFE HARBOR MATCH, PROFIT SHARING , SAFE HARBOR NEC .

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,268 @@
...
+                        "[ReportPayrollContribution] Not able to write the row for employee: {} due to error {}".format(
+                            ssn, e
+                        )
+                    )
+                    raise Exception(
+                        "[ReportPayrollContribution] Not able to write the row for employee: {}".format(
+                            ssn
+                        )
+                    )
+            file = File()
```
vguptarippling, line 261: Doesn't Vanguard/Ascensus file have file header ? Do we have any new file spec or we are using old Vanguard code  here?

codewizzy: We are using old vanguard code here.

flux_sdk/pension/test_utils/test_ascensus/test_ascensus.py
codewizzy: Tried freeze_time but it didn't worked out due to import error
`flux_sdk/pension/utils/tests/test_ascensus/test_ascensus.py:21: error: Cannot find implementation or library stub for module named "freezegun"  [import-not-found]`
Moreover, we are testing it other files same way and it is working fine.
Ex `submodules/flux-apps/flux_apps/charlesschwab/pension_kit/capabilities/report_payroll_contributions/test/test_implementation.py:99`

flux_sdk/pension/utils/ascensus.py
vguptarippling: Then we don't we have file header like we have in Vanguard. Also, where is site_code, site_code_mappings ?

flux_sdk/pension/utils/ascensus.py
pshanmukrao: I am worried that we have a hard dependency on where the lambda is running as it takes local timezone.

flux_sdk/pension/utils/tests/test_ascensus/test_ascensus.py
```diff
@@ -0,0 +1,171 @@
...
+        ]
+        incomplete_employee_payroll_records = self.employee_payroll_records
+        for required_field in required_employee_payroll_records_information:
+            for index, incomplete_employee_payroll_record in enumerate(
+                incomplete_employee_payroll_records
+            ):
+                self.set_nested_attribute(
+                    incomplete_employee_payroll_record, required_field, None
+                )
+            with self.assertRaises(Exception):
```
pshanmukrao, line 130: is there a specific Exception that we can assert ?

pshanmukrao (approved):

flux_sdk/pension/utils/ascensus.py
codewizzy: We don't need "DIVISIONAL CODE" as input from customer as it is not mandatory.

flux_sdk/pension/utils/ascensus.py
codewizzy: Done

flux_sdk/pension/utils/ascensus.py
codewizzy: Have to add these.

flux_sdk/pension/utils/tests/test_ascensus/test_ascensus.py
codewizzy: We are simply raising exception if data is missing.

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+    For further details regarding their implementation details, check their documentation.
+    """
+
+    @staticmethod
+    def get_formatted_ssn(ssn: str) -> str:
+        return ssn[:3] + "-" + ssn[3:5] + "-" + ssn[5:9]
+
+    @staticmethod
+    def to_bytes(content: Union[str, bytes]) -> bytes:
+        data = content.encode() if isinstance(content, str) else content
```
Kingpin007, line 88:
```suggestion
        return content.encode() if isinstance(content, str) else content
```

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+    Developers are required to implement the format_contributions_for_ascensus_vendor method in their code.
+    For further details regarding their implementation details, check their documentation.
+    """
+
+    @staticmethod
+    def get_formatted_ssn(ssn: str) -> str:
+        return ssn[:3] + "-" + ssn[3:5] + "-" + ssn[5:9]
+
+    @staticmethod
+    def to_bytes(content: Union[str, bytes]) -> bytes:
```
Kingpin007, line 87: This is python 3.10, typing can be `str | bytes`

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+        data = content.encode() if isinstance(content, str) else content
+        return data
+
+    @staticmethod
+    def get_amount_from_payroll_contribution(
+        payroll_contribution: Optional[PayrollRunContribution],
```
Kingpin007, line 93: typing can be `PayrollRunContribution | None`

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+        )
+        payroll_run_id = payroll_upload_settings.payrun_info.payroll_run_id
+
+        if not environment or not client_id:
+            raise RuntimeError(
+                "Environment and client_id must be present to upload Vanguard file"
+            )
+        timestamp = datetime.now()
+        format_timestamp = timestamp.strftime("%m%d%Y.%H%M%S")
+        return "{}_{}_{}_{}.csv".format(
```
Kingpin007, line 121: can we use f-string here?

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+        client_id = str(
+            payroll_upload_settings.customer_partner_settings[
+                AscensusSettingsKeys.CLIENT_ID.value
+            ]
+        )
+        payroll_run_id = payroll_upload_settings.payrun_info.payroll_run_id
+
+        if not environment or not client_id:
+            raise RuntimeError(
+                "Environment and client_id must be present to upload Vanguard file"
```
Kingpin007, line 117: Can we add better error message here, we can print values of Environment and client_id so we know which one is missing

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+        current_payroll_run_id = payroll_upload_settings.payrun_info.payroll_run_id
+        sorted_payroll_run_ids = sorted(
+            [payroll_run.payroll_run_id for payroll_run in current_month_payroll_runs]
+        )
+        return sorted_payroll_run_ids.index(current_payroll_run_id)
+
+    @staticmethod
+    def _get_fein_site_code(
+        pay_frequency: Optional[str], pay_type: Optional[str], fein_settings: dict
+    ) -> str:
```
Kingpin007, line 172: return type seems incorrect it should be `str | None` look at the last line int the function.

@stevevls/ @ujjwalshukla / @vguptarippling   I think there may be some issue with the mypy setup for this repo.

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+            return fein_settings[fein_site_code_frequency_mapping_key]
+        if fein_site_code_pay_type_mapping_key in fein_settings:
+            return fein_settings[fein_site_code_pay_type_mapping_key]
+
+        return fein_settings.get("fein_site_code", None)
+
+    @staticmethod
+    def _get_default_site_code(
+        preference_type: str,
+        pay_frequency: Optional[str],
```
Kingpin007, line 191: Generally optional arguments should not be followed by a required argument. Please re-order them.

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+
+    @staticmethod
+    def get_file_name(payroll_upload_settings: PayrollUploadSettings) -> str:
+        """
+        Receives a PayrollUploadSettings from which the developer
+        is expected to return file name based on payroll_upload_settings
+        :param payroll_upload_settings:
+        :return: str
+        """
+        environment = payroll_upload_settings.environment
```
vguptarippling, line 105: You need to take env from customer_partner_settings. This field should be removed soon. Please use the same as paykonnect

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+    """
+    This class embodies the functionality to "report payroll contributions" for vendors utilizing
+    the Ascensus formatted file.
+    Developers are required to implement the format_contributions_for_ascensus_vendor method in their code.
+    For further details regarding their implementation details, check their documentation.
+    """
+
+    @staticmethod
+    def get_formatted_ssn(ssn: str) -> str:
+        return ssn[:3] + "-" + ssn[3:5] + "-" + ssn[5:9]
```
vguptarippling, line 83: This will remove the last character from ssn string right? Can you make ssn[5:]

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+
+    @staticmethod
+    def _get_preference_type_site_code(
+        preference_type: str, fein_settings: dict, customer_partner_settings: dict
+    ) -> str:
+        fein_site_code = fein_settings.get("fein_site_code", None)
+        if fein_site_code:
+            return fein_site_code
+        return customer_partner_settings.get(
+            f"site_code_{preference_type.lower()}", "A"
```
vguptarippling, line 219: Shouldn't this be default_site_code in worst case?

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+
+        if pay_frequency is None and pay_type is None:
+            return (
+                ReportPayrollContributionsAscensusUtil._get_preference_type_site_code(
+                    preference_type, fein_settings, customer_partner_settings
+                )
+            )
+
+        fein_site_code_mapping_key = (
+            f"fein_site_code_mapping_for_{preference_type.lower()}"
```
vguptarippling, line 266: This only contains the fein selections for the preference type right? returning this on next would return the value not site code right ?

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+        pay_frequency: str | None,
+        pay_type: str | None,
+    ) -> str:
+        site_code_frequency_mapping_key = (
+            f"site_code_mapping_for_{pay_frequency.lower()}" if pay_frequency else None
+        )
+        site_code_pay_type_mapping_key = (
+            f"site_code_mapping_for_{pay_type.lower()}" if pay_type else None
+        )
+        if site_code_frequency_mapping_key in customer_partner_settings:
```
vguptarippling, line 201: add null check site_code_frequency_mapping_key and  site_code_frequency_mapping_key in customer_partner_settings:

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+            return customer_partner_settings[site_code_frequency_mapping_key]
+        if site_code_pay_type_mapping_key in customer_partner_settings:
```
vguptarippling, line 203: add null check

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+    ) -> str | None:
+        fein_site_code_frequency_mapping_key = (
+            f"fein_site_code_mapping_for_{pay_frequency.lower()}"
+            if pay_frequency
+            else None
+        )
+        fein_site_code_pay_type_mapping_key = (
+            f"fein_site_code_mapping_for_{pay_type.lower()}" if pay_type else None
+        )
+        if fein_site_code_frequency_mapping_key in fein_settings:
```
vguptarippling, line 181: add null check

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+            return fein_settings[fein_site_code_frequency_mapping_key]
+        if fein_site_code_pay_type_mapping_key in fein_settings:
```
vguptarippling, line 183: add null check

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,456 @@
...
+
+    @staticmethod
+    def get_index_of_current_payroll_run(
+        payroll_upload_settings: PayrollUploadSettings,
+    ) -> int:
+        current_month_payroll_runs = getattr(
+            payroll_upload_settings, "current_month_payruns", []
+        )
+        current_payroll_run_id = payroll_upload_settings.payrun_info.payroll_run_id
+        sorted_payroll_run_ids = sorted(
```
vguptarippling, line 164: filter them out by check date

flux_sdk/pension/utils/ascensus.py
codewizzy: Yes, nice catch.

flux_sdk/pension/utils/ascensus.py
codewizzy: It is, check `_get_default_site_code`.
If pay_type and pay_frequency both are none then their is no point of checking on basis of site code mapping that is why returning here, but there might be case if pay_type or pay_frequency is present still no site code mapping exists.

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,453 @@
...
+        site_code = customer_partner_settings.get(
+            f"site_code_{preference_type.lower()}", "A"
+        )
+        return site_code
+
+    @staticmethod
+    def _get_preference_type_site_code(
+        preference_type: str, fein_settings: dict, customer_partner_settings: dict
+    ) -> str:
+        fein_site_code = fein_settings.get("fein_site_code", None)
```
vguptarippling, line 218: I think this should be default to site_code if pay_frequency and pay_type is None as written in old codebase

flux_sdk/pension/utils/ascensus.py
```diff
@@ -0,0 +1,453 @@
...
+    @staticmethod
+    def get_site_code(
+        fein_settings: dict,
+        payroll_update_settings: PayrollUploadSettings,
+        no_of_salaried_roles: int,
+        no_of_hourly_roles: int,
+    ) -> str:
+        """
+        This method is used to get the site code for the given payroll run
+        1. It tries to get the site code from the fein settings for preferred payroll frequency type
```
vguptarippling, line 234: Can we update the comment in order like ?
First check if both of them is none then default
Then fein settings preferrable else default
Then generic setting preferrable else default
as it is slightly confusing

flux_sdk/pension/utils/ascensus.py
codewizzy: We fetch site_code using `get_fein_or_settings` method in old code, in case `site_code` is not in fein setting we fallback to default site_code

flux_sdk/pension/utils/ascensus.py
codewizzy: Sure

vguptarippling (approved):

PR #31, opened 2023-12-18
Add ETL Kit (Data Import)

dominicbarnes:
This PR adds a new "ETL" kit which is to be used for Data Import. There is 1 capability called "Single Object Import" which has support for MongoDB and SQL queries as data sources.

This kit uses Python dataclasses instead of vanilla classes for the data models, and includes validation via `__post_init__` rather than an external library/dependency.

This also tweaks the lint workflow to use Poetry for managing dependencies.

Previous discussion available at #23 

Lastly, this bumps the package minor version to v0.10 as it is the addition of a new kit altogether.


flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,61 @@
...
+from flux_sdk.etl.data_models.query import Connector, Query
+from flux_sdk.etl.data_models.record import Checkpoint, Record
+from flux_sdk.etl.data_models.schema import Schema
+
+
+class SingleObjectImport(ABC):
+    """
+    This class represents the "single_object_import" capability. This is the first iteration of our Data Import
+    framework, but is limited in scope to only operating on a single object at a time. This is a stepping stone towards
+    a more robust system, so consider this experimental.
```
stevevls, line 13: Sorry for not providing this feedback before, but let's talk about what it does rather than future states.  I don't think we should talk about it being experimental either.

flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,61 @@
...
+    @staticmethod
+    @abstractmethod
+    def process_records(schema: Schema, records: list[Record]) -> list[Record]:
+        """
+        Use this hook to post-process the Records extracted via the query performed from "prepare_query". This hook is
+        required in order to implement incremental sync (via checkpoint) or to dynamically drop records from the sync
+        (via drop).
+
+        If the Record does not adhere to the provided schema, such as using an incorrect/incompatible type for a field,
+        the sync will fail. As such, this hook is the opportunity to massage any raw query results that may not be
```
stevevls, line 55:
```suggestion
        the sync will fail. As such, this hook has the opportunity to massage any raw query results that may not be
```

flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,61 @@
...
+        directly compatible.
+
+        :param schema: The schema generated in the "get_schema" hook for this object.
+        :param records: This batch of records should be updated
```
stevevls, line 59:
```suggestion
        :param records: The batch of records to be updated
```
?

flux_sdk/etl/capabilities/single_object_import/interface.py
```diff
@@ -0,0 +1,61 @@
...
+        :return: Query
```
stevevls, line 60: This should be records, no?

flux_sdk/etl/data_models/query.py
```diff
@@ -0,0 +1,73 @@
+from dataclasses import dataclass
```
stevevls, line 1: Should comments in this file be docstrings?

flux_sdk/etl/data_models/query.py
```diff
@@ -0,0 +1,73 @@
...
+    args: Optional[dict[str, SQLQueryArg]] = None
+
+    # Perform validation.
+    def __post_init__(self):
+        if not self.text:
+            raise ValueError("text is required")
+        elif not isinstance(self.text, str):
+            raise TypeError("text must be a string")
+
+        if self.args:
```
stevevls, line 37: Do we want to check that the args are valid types as well?

flux_sdk/etl/data_models/query.py
```diff
@@ -0,0 +1,73 @@
...
+        elif not isinstance(self.collection, str):
+            raise TypeError("collection must be a string")
+
+        if self.filter:
+            if not isinstance(self.filter, dict):
+                raise TypeError("filter must be a dict")
+
+        if self.aggregate:
+            if not isinstance(self.aggregate, list):
+                raise TypeError("aggregate must be a list")
```
stevevls, line 69: Do we want to check the list members to be dicts?

flux_sdk/etl/data_models/record.py
```diff
@@ -0,0 +1,63 @@
+from dataclasses import dataclass
```
stevevls, line 1: Ditto re: docstrings

flux_sdk/etl/data_models/record.py
```diff
@@ -0,0 +1,63 @@
...
+# Checkpoints can be represented as either timestamp (datetime), sequence number (int) or an opaque token (str). Values
+# are compared assuming an increasing sort order, so the largest value in a sync will be treated as the high-water mark
+# for the next sync.
+Checkpoint = Union[datetime, int, str]
+
+
+# This corresponds to a row in the source database.
+@dataclass(kw_only=True)
+class Record:
+    # The ID for this record, which becomes external_id in the Custom Object. Since only a simple, string value is
```
stevevls, line 18: Let's tie this comment in to the field name to be explicit about the fact that this is the primary key in the customer's system of record

flux_sdk/etl/data_models/record.py
```diff
@@ -0,0 +1,63 @@
...
+    # supported by Custom Objects, this does not need to be a full SchemaField.
+    primary_key: str
+
+    # This is the rest of the raw data for this record.
+    fields: dict[str, Field]
+
+    # This represents the links in this object. Each key is the field name, which is used to look up the target details.
+    # Each value is the identifier used for the link.
```
stevevls, line 26: Should we add an example in the comment?

flux_sdk/etl/data_models/record.py
```diff
@@ -0,0 +1,63 @@
...
+    # Perform validation.
+    def __post_init__(self):
+        if not self.primary_key:
+            raise ValueError("primary_key is required")
+        elif not isinstance(self.primary_key, str):
+            raise TypeError("primary_key must be a string")
+
+        if not self.fields:
+            raise ValueError("fields is required")
+        elif not isinstance(self.fields, dict):
```
stevevls, line 50: Do we need to check the dict key/values to ensure they're strings?

flux_sdk/etl/data_models/record.py
```diff
@@ -0,0 +1,63 @@
...
+            raise TypeError("fields must be a dict")
+
+        if self.references:
+            if not isinstance(self.references, dict):
+                raise TypeError("references must be a dict")
+
+        if self.checkpoint:
+            if type(self.checkpoint) not in get_args(Checkpoint):
+                raise TypeError("checkpoint must be a datetime")
```
stevevls, line 59: Error message is incorrect?

flux_sdk/etl/data_models/record.py
```diff
@@ -0,0 +1,63 @@
...
+    # after the sync is complete. Rippling will keep track of the *highest* value seen from all records. That value will
+    # be passed to the "prepare_query" hook and can be used to adjust the query to only retrieve the records that have
+    # changed since the last sync.
+    #
+    # This must be used to enable incremental sync, otherwise a full sync will happen each time.
+    checkpoint: Optional[Checkpoint] = None
+
+    # This flag can be used by the "process_records" hook to signal to Rippling that the object should not be imported.
+    # If a Record is not found after this hook, that will be regarded as an error, so this flag should be used instead.
+    drop: Optional[bool] = None
```
stevevls, line 39: WDYT about having this be a `bool` with a default value of `False`?  Making it optional turns it into a ternary state, but I'm unsure why someone would ever set `drop` to false

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,238 @@
+from dataclasses import dataclass
```
stevevls, line 1: Docstrings for this file too

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,238 @@
...
+
+        if not isinstance(self.is_required, bool):
+            raise TypeError("is_required must be a bool")
+
+        if not isinstance(self.is_unique, bool):
+            raise TypeError("is_unique must be a bool")
+
+        if self.enum_values:
+            if not isinstance(self.enum_values, list):
+                raise TypeError("enum_values must be a list")
```
stevevls, line 73: Do we need to check type of list values?

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,238 @@
...
+
+# A schema describes the shape of an object being imported and is used by Rippling to define the Custom Object.
+@dataclass(kw_only=True)
+class Schema:
+    # This is the name of the Custom Object.
+    name: str
+
+    # This is used to put this Custom Object into a group/category. If a category with this name already exists, a new
+    # one will not be created.
+    category_name: str
```
stevevls, line 157: What's the behavior if this changes?

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,238 @@
...
+
+    # This is the name of the field used for the record name, which is displayed in the Rippling Custom Object UI. When
+    # not provided,
+    name_field: str
+
+    # These are the remaining data fields.
+    fields: list[SchemaField]
+
+    # This is the name of the field used for the primary key (aka: External ID).
+    primary_key_field: Optional[str] = None
```
stevevls, line 171: Why is this one optional?

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,238 @@
...
+
+        if not self.category_name:
+            raise ValueError("category_name is required")
+        elif not isinstance(self.category_name, str):
+            raise TypeError("category_name must be a str")
+
+        if not self.category_description:
+            raise ValueError("category_description is required")
+        elif not isinstance(self.category_description, str):
+            raise TypeError("category_description must be a str")
```
stevevls, line 205: I keep seeing blocks like this over and over and wonder if we should decomp out a couple of helpers.  Like `raise_if_missing_or_wrong_type("category_description")` that could `getattr` on the class and do the checks.

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,238 @@
...
+            if not isinstance(self.last_modified_date_field, str):
+                raise TypeError("last_modified_date_field must be a str")
+
+        if self.references:
+            if not isinstance(self.references, dict):
+                raise TypeError("references must be a dict")
+
+        if self.owner:
+            if not isinstance(self.owner, tuple):
+                raise TypeError("owner must be a tuple")
```
stevevls, line 238: Do we need next level down checks?

stevevls (commented):
Yay for unit tests. :)

flux_sdk/etl/data_models/query.py
dominicbarnes: It's a weird pattern to me too, I'm just following the convention of the other kits at this point.

stevevls: Hmm...weird indeed.  I think we should do real documentation if possible.  Unless someone knows a reason not to? 🤔

flux_sdk/etl/data_models/schema.py
dominicbarnes: @sz2376 informed me that this isn't actually required by Custom Objects.

stevevls: Woah.  What does it mean if we don't have a PK?  Should _we_ require it even if CO doesn't?

flux_sdk/etl/data_models/query.py
dominicbarnes: I've raised this in the Flux channel, let's see if we can get some feedback or context we don't currently have.

flux_sdk/etl/data_models/record.py
dominicbarnes: I'm open to this, I left it this way because I was basically considering this to only be set by the "process_records" hook, and I wouldn't want the default `False` to show up in other contexts and cause confusion, as `None` would definitely be distinct.

flux_sdk/etl/data_models/schema.py
dominicbarnes: It sounds like CO creates a PK automatically, which now that I think about is kind of wild as we would need to keep track of this somewhere otherwise we'd just be creating new records every time we sync.

cc/ @sz2376 we probably need to reconsider requiring this so we can use the same value for both the ID and ExternalID.

flux_sdk/etl/data_models/schema.py
dominicbarnes: I did some drying up of this code.

flux_sdk/etl/data_models/schema.py
stevevls: My two cents would be that we should just require the column name in this particular context.  We can impose additional constraints to CO where they make sense for the use case.

sz2376: I think we can require this for now(even though it's optional in CO) otherwise we might end up creating duplicate records in CO; I am not sure whether there is any use case where people do want to create new records every time we import the data(most likely not).

flux_sdk/etl/data_models/schema.py
```diff
@@ -0,0 +1,186 @@
...
+    is_required: bool = False
+    """When enabled, this value will be required to be not empty by Rippling Custom Object."""
+
+    is_unique: bool = False
+    """When enabled, this value will be enforced as unique by Rippling Custom Object."""
+
+    enum_values: Optional[list[str]] = None
+    """When using an Enum data type, this indicates the possible values."""
+
+    enum_restricted: bool = False
```
sz2376, line 48: This should be optional

sz2376 (approved):

flux_sdk/etl/data_models/schema.py
sz2376: If the category_name gets changed in the app implementation, we will create a new category.

sz2376 (commented):
BTW, shall we also bump up the version?

flux_sdk/etl/data_models/schema.py
dominicbarnes: It is optional, the default is `False`

stevevls (approved):
:shipit:

PR #37, opened 2023-12-21
Add FSA plan type

fanchen1988:
https://rippling.atlassian.net/browse/INSURANCE-68090

### DIFF
--- a/flux_sdk/flex/capabilities/update_contributions/data_models.py
+++ b/flux_sdk/flex/capabilities/update_contributions/data_models.py
@@ -3,6 +3,7 @@ from decimal import Decimal
 from enum import Enum
 from typing import Optional
 
+from flux_sdk.flex.data_models.flex_models import FSAPlanType
 from flux_sdk.flux_core.data_models import Employee
 
 
@@ -24,3 +25,4 @@ class EmployeeContribution:
     contribution_amount: Decimal
     tax_year: Optional[str]
     employee: Optional[Employee]
+    fsa_plan_type: Optional[FSAPlanType]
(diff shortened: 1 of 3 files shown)


flux_sdk/flex/capabilities/data_models.py
```diff
@@ -0,0 +1,8 @@
+from enum import Enum
+
+
+# specify the plan type for FSA
+class FSAPlanType(Enum):
```
vguptarippling, line 5: I am not sure if we are following this convention to add data_models inside capabilities @pshanmukrao @stevevls thoughts on this ?

fanchen1988: This is for transferring capabilities data, and both enrollment and contribution will use it

Kingpin007 (approved):

flux_sdk/flex/capabilities/data_models.py
pshanmukrao: We probably want Kit level data_models in that case.

See: https://github.com/Rippling/flux-sdk/blob/main/flux_sdk/etl/data_models/schema.py

fanchen1988: It's not the Kit level, only FSA could use it @pshanmukrao

pshanmukrao: Both capabilities in a Kit can use it right ? Then isn't it a "Kit" level model? 

Can you help me what you mean by only FSA could use it ?

fanchen1988: Only Flex will use it, it's a Flex benefit specific data model

pshanmukrao: in that case add it in 

e.g., flux_sdk/flex/data_models/flex_models.py

fanchen1988: @pshanmukrao done

pshanmukrao (approved):

PR #40, opened 2023-12-29
Update aggregate type from Any to dict<str, Any>

pshanmukrao:
https://rippling.atlassian.net/browse/APPS-23936

Currently FluxDecoder doesn't support Any field , but does support dict<str, Any>

### DIFF
--- a/flux_sdk/etl/data_models/query.py
+++ b/flux_sdk/etl/data_models/query.py
@@ -66,7 +66,7 @@ class MongoQuery:
     This provides the projection criteria for a basic find operation. Use this to include/exclude fields at query-time.
     """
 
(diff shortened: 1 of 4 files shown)


falcon2212 (approved):

flux_sdk/etl/data_models/query.py
```diff
@@ -66,7 +66,7 @@ class MongoQuery:
     This provides the projection criteria for a basic find operation. Use this to include/exclude fields at query-time.
     """
 
-    aggregate: Optional[list[Any]] = None
+    aggregate: Optional[list[dict[str, Any]]] = None
```
Kingpin007, line 69: for future reference lets not use Optional anymore we can replace with (type) | None

PR #48, opened 2024-02-14
ETL Kit: use typing.NewType for type aliases so typing is retained at runtime

stevevls:
This PR updates the syntax for creating type aliases such that the variable type names are preserved at runtime.

Type alias information is lost at runtime when using the more standard type alias syntax like:

```
SQLQueryArg = Union[str, bool, int, float, datetime, date, time]
```

Tools that use introspection will not see type annotations of `SQLQueryArg` but will see the union type that has been aliased instead.

Using a different syntax, we can ensure that introspection-based tools can still see type per the original type declaration.
```
SQLQueryArg = NewType("SQLQueryArg", Union[str, bool, int, float, datetime, date, time])
```

Validation and tests were updated because now there is an extra step required to go from type alias to the concrete type definition.

### DIFF
--- a/flux_sdk/etl/data_models/schema.py
+++ b/flux_sdk/etl/data_models/schema.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
-from typing import Optional, Union
+from typing import NewType, Optional, Union
 
 from flux_sdk.flux_core.validation import check_field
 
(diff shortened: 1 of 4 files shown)


flux_sdk/etl/data_models/schema.py
pshanmukrao, line 1: is there a way we can prevent future Kit developers to not write Union ? may be some tests ?

stevevls:
Abandoning this PR because I believe we've solved it in the scaffolding codegen and no longer have special requirements on how type aliases are declared.

PR #57, opened 2024-04-04
Update python version to 3.10 to support dataclasses

pshanmukrao:


stevevls (approved):
LGTM once you get the poetry lock regenerated.  As a sanity check, this won't affect any of the deployed lambda functions?  Just wanting to make sure that we've thought through any migration scenarios. 😅

pshanmukrao:
> LGTM once you get the poetry lock regenerated. As a sanity check, this won't affect any of the deployed lambda functions? Just wanting to make sure that we've thought through any migration scenarios. 😅

Surprisingly, our flux-sdk is currently deployed under python 3.11 runtime. :-D

sz2376 (approved):

PR #73, opened 2024-06-07
Add group_number to bswift EE model

bryanbierce:


vguptarippling (approved):

sz2376 (approved):
approving to unblock(AFAIS there is already group_number field used in this file so it sounds to be of the right name and type) but as we discussed offline please add the doc string for each individual field as well which is critical for doc generation.

PR #84, opened 2024-09-18
A lightweight user management kit definitions

sz2376:
A lightweight user management kit definitions. The doc strings should explain it clearly but if not please leave comments so that we can make them self sufficient.


flux_sdk/flux_core/data_models.py
```diff
@@ -1,8 +1,40 @@
...
+
+class BaseHttpRequest:
+    """The base class for all HTTP request data classes."""
+
+    next_page: Optional[str] = None
+    """
+    The next page token to fetch the next page of results.
+    The initial request will always have this field as None;
+    when the previous response has the next_page set, the same hook will be invoked again with the next_page set in the request.
+    """
```
stevevls, line 27: I don't think we should have the protocol in the name here...it doesn't match the way we've been thinking about connectors so far.  IIUC, I think what these are meaning to convey is pagination moreso than HTTP.  So maybe we can change the name?

flux_sdk/user_management/capabilities/user_lifecycle_management/interface.py
```diff
@@ -0,0 +1,35 @@
...
+
+class UserLifecycleManagement(BaseCapability):
+    """
+    This class represents the "user_lifecycle_management" capability. This capability is used to manage the lifecycle
+    of users in the third-party system with supports including:
+    - Getting user's organizations from the third-party system
+    - Getting users from the third-party system
+    """
+
+    def get_users(self, request: GetUsersRequest) -> GetUsersResponse:
```
stevevls, line 18: Could this be simpler if we didn't paginate?  Do we expect responses so large that we wouldn't want to just get all users at once?

sz2376: The app developer does not necessarily need to make this pagination but we are more like providing such support for them in case there is a large payload to be returned as a whole(exceeding the 6 MB size limit for Lambda) or it takes too much time to fetch all the users(exceeding the 15 min latency limit but we could have set a smaller timeout in our current implementation). I do agree it's very less likely the `get_users` hook would exceed such limits but there might be more costly hooks in the future? I am fine if we want to remove the pagination support for now and add it in the future if we do see such use case arise.

flux_sdk/flux_core/data_models.py
sz2376: I agree that this is not tied up with the HTTP protocol at all and we just happen to have the pagination scenario for the http use case. What about other names like `BaseOutboundRequest`, `BaseExternalRequest`, `BaseAppRequest` or just `BaseRequest`? Essentially the BaseRequest or BaseResponse here is meant to be used by the app developer to make external calls to their endpoint and return us the results, plus indicating whether the app is disconnected requiring the customer to reconnect. In the future we might want to add more fields into the response so that the app developer could provide more info to us, for example, like the app is missing some particular scopes so the request did not go through.

flux_sdk/flux_core/data_models.py
```diff
@@ -1,8 +1,30 @@
...
+    """The base class for all external request made to the third party system."""
+    pass
+
+
+class BaseCapability(ABC):
+    """The base class for all capabilities."""
+
+    def __init__(self, session: requests.Session | None = None):
+        self.session = session
+        """The session to make HTTP requests to the third party."""
```
stevevls, line 25: Does this need to be injected?  IIRC, the plan was to use environment variables for proxy config that would be automatically picked up by urllib or requests.  That way, we don't need to force the app author to use a particular package and they can use whatever they like best.

flux_sdk/flux_core/data_models.py
```diff
@@ -1,8 +1,30 @@
...
 
 
+class BaseExternalResponse:
+    """The base class for all the response from the third party system."""
+
+    is_app_disconnected: bool = False
+    """This indicates if the app is disconnected from the third-party system."""
+
+
+class BaseExternalRequest:
```
stevevls, line 15: If there's nothing in here, can we drop it and add it in later if we find a need?

flux_sdk/flux_core/data_models.py
```diff
@@ -1,8 +1,30 @@
...
 from datetime import date, datetime
 from enum import Enum
+import requests
 from typing import Optional
 
 
+class BaseExternalResponse:
+    """The base class for all the response from the third party system."""
+
+    is_app_disconnected: bool = False
```
stevevls, line 11: Is this better captured as a named exception?

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@dataclass(kw_only=True)
+class Organization:
+    """
+    This represents an organization in the third party system. An organization is used to group users in the third party system.
+    Different third party systems may have different names of organizations: Github calls it "Organization", while Google Workspace calls it "Domain".
+    """
+
+    id: str
```
stevevls, line 13: Is this an external ID or a rippling ID?  Should we have both or just one?

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
```diff
@@ -0,0 +1,68 @@
...
+
+    name: Optional[str] = None
+    """The name of the organization in the third party system."""
+
+
+@dataclass(kw_only=True)
+class User:
+    """This represents a user in the third party system."""
+
+    id: Optional[str] = None
```
stevevls, line 24: Is this optional because sometimes we're using the class to create a user?  If so, should we document that the id must be populated for any read operation from the SoR?

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
```diff
@@ -0,0 +1,68 @@
...
+
+    id: str
+    """The unique identifier of the organization in the third party system."""
+
+    name: Optional[str] = None
+    """The name of the organization in the third party system."""
+
+
+@dataclass(kw_only=True)
+class User:
```
stevevls, line 21: Is there a relationship between user and org that should be captured?  For example, if I were to to call `GetUsersRequest` without specifying orgs, should I be able to pick apart the resulting slice by org?

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
sz2376: This is just the external ID; the third party does not know the rippling ID.

flux_sdk/flux_core/data_models.py
sz2376: We need this injection when we are invoking the app implementation in ripplina-main cause we cannot set a environment variable for the whole rippling-main to affect all the requests. For the Lambda invocation, I am inclined to set the proxy by the environment variable so that the app developer could use whatever urllib they want. I think we briefly discussed about this but we did not figure out a very elegant approach when invoking the hooks in rippling-main comparing to the Lambda approach; please let me know if there is a better way around.

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
sz2376: I think we have mentioned that in the comment but please let me know if you want to reword it.

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
sz2376: In our current use cases, the organizations are mostly used when fetching the users and that's why we have the organizations in the GetUsersRequest; I don't see a use case where we want to filter the users by the organizations they belong to. Even if there is such use case, it should still be the app developer doing such filtering rather than us, right? If we do see the use case where we really want to do such filtering by ourselves, we can just add one more optional field in the User class to indicate the organization(s) of the user. One thing I am not too sure is whether it should be a str or a list of str but when the real use case comes we would have better idea about it. WDYT?

flux_sdk/flux_core/data_models.py
sz2376: That also works; then I think we can get rid of both BaseExternalRequest and BaseExternalResponse for now.

stevevls: I like getting rid of stuff. :)

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
stevevls: Cool.  Let's document that?  I think you documented another ID in this respect.

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
stevevls: Which comment?  Maybe I missed it...

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
stevevls: Sure, if we can punt for now that works.  My guess is that it would be a collection, but better to know later than to guess now!

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
sz2376: > The unique identifier of the organization in the third party system.

I thought this was sufficient, right?

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
sz2376: > The unique identifier of the user in the third party system. This field is required when getting users from the third party system.

This was what I had but please let me know if you want to reword it.

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
sz2376: Yeah I am not sure if it's a collection or a dictionary where the app developer can tell us the `default` organization in case a user can belong to multiple organization while having one as the default/primary.

flux_sdk/flux_core/data_models.py
stevevls: Here's an idea.  Could you have a mixin that's internal to RM that your internal kit can use to inject this?  I think injecting the session makes sense, but I still am not convinced that it's part of the public contract.

sz2376: I like this idea! I think we can change the constructor in the runtime when we instantiate the capability class to ingest in the session as we want.

flux_sdk/user_management/capabilities/user_lifecycle_management/data_models.py
stevevls: I'm sorry...I forgot how to read pydoc. 🤦  I was expecting it above the method.

stevevls (approved):

sihan416 (approved):

PR #88, opened 2024-09-30
[APPS-29594] add get job attributes sync sdk

dhshah-rippling:
We allow 3P developers to define a mapping between their attribute and all the values their system supports.

We changed the initial approach that instead of having two hooks, we now have one hook to get this data.

Example responses:

https://docs.google.com/document/d/1MlJkZluGtTAiYQ0rSNyi8Y9elfumDskqRSFg7E1rEJU/edit

image


flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -44,6 +44,159 @@ def __post_init__(self):
...
+    """The city of the address."""
+
+    state: str
+    """The state of the address."""
+
+    country_code: str
+    """The 2 letter uppercase country code of the address."""
+
+    phone: str
+    """The phone number of the address."""
```
sz2376, line 73: It's a little surprising to see the phone to be part of the address, especially as a required field. Could you clarify why this is needed?

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -44,6 +44,159 @@ def __post_init__(self):
...
+        check_field(self, "phone", str, required=True)
+
+@dataclass(kw_only=True)
+class PayRateCompatibleValue:
+    """
+    This represents a pay rate compatible value.
+    """
+
+    job_title: str
+    """The job title associated with the pay rate."""
```
sz2376, line 96: This should be a generic field called `name` indicating the name of the attribute value:
`name: "Cashier"` for example.

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -44,6 +44,159 @@ def __post_init__(self):
...
+
+    pay_rate: Optional[float] = None
```
sz2376, line 98: let's use decimal type for better accuracy

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -44,6 +44,159 @@ def __post_init__(self):
...
+
+    def __post_init__(self):
+        """Perform validation."""
+        check_field(self, "job_title", str, required=True)
+
+        if self.pay_rate is not None and not isinstance(self.pay_rate, float):
+            raise ValueError(f"pay_rate must be a float if provided, got {type(self.pay_rate).__name__} instead.")
+
+@dataclass(kw_only=True)
+class JobSiteLocationCompatibleValue:
```
sz2376, line 109: We don't have the name for this value? I wonder how the FE looks like if there is no name for this job code.

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -44,6 +44,159 @@ def __post_init__(self):
...
+class WorkLocationCompatibleValue:
+    """
+    This represents a work location compatible value.
+    """
+
+    name: str
+    """The name of the work location."""
+
+    active: bool
+    """Whether the work location is enabled or not. This is not the same as whether it is closed for the day."""
```
sz2376, line 131: What's the necessity for the app developer to specify whether this value is active or not? I know we have this boolean in our internal work location record but that might be different concept of whether the store location is active or not. Let's get rid of this for now unless it's really necessary to sync this field from the third party?

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -44,6 +44,159 @@ def __post_init__(self):
...
+
+    latitude: float
+    """The latitude of the work location."""
+
+    longitude: float
+    """The longitude of the work location."""
```
sz2376, line 137: I feel some third parties might not have such info available so let's either get rid of them for now or at least mark them as optional

flux_sdk/time_and_attendance/capabilities/job_management/interface.py
```diff
@@ -15,3 +18,14 @@ def get_job_attributes(self) -> GetJobAttributesResponse:
         :return: The response containing the job attributes from the third-party system.
         """
 
+    @abstractmethod
+    def get_job_attributes_values(self) -> GetJobAttributesValueResponse:
```
sz2376, line 22:
```suggestion
    def get_job_attribute_values(self) -> GetJobAttributeValuesResponse:
```
This might be a naming preference for myself but just try to keep it consistent with the `get_job_attributes` api as well.

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -44,6 +44,159 @@ def __post_init__(self):
...
+    """
+    This represents a response containing the attributes from the third party mapped to their most compatible values.
+
+    For instance, if the third party system has an attribute "work_location" that is most compatible with Rippling's
+     "WORK_LOCATION" attribute, and something like the following is defined:
+
+    result = {
+        "work_location": [WorkLocationCompatibleValue(name="sf",...), WorkLocationCompatibleValue(...)],
+        "pay_rate": [PayRateCompatibleValue(job_title="cashier",...)],
+        "job_title": [PayRateCompatibleValue(...)],
```
sz2376, line 161: In this dictionary, the key would be the attribute in the third party, which is what returned from the get_job_attributes hook. For QuBeyond example it would be `job_title` and `location`.

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: talked w t&a team and they said its not needed on their side, so will remove

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: sure

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: I guess its not necessary. I thought it might be helpful in payroll somehow -- like are they supposed to pay employees belonging to a closed location?

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: we can get rid of them. its hard to understand what the team will use or wont use because all these fields on address are optional, but i think its safe to remove them

flux_sdk/time_and_attendance/capabilities/job_management/interface.py
dhshah-rippling: yeah i think this is a good change

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: yeah here i put as example, since other companies may use pay_rate key .

I will clarify in comments

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: fixed

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -44,6 +45,158 @@ def __post_init__(self):
...
+
+@dataclass(kw_only=True)
+class PayRateCompatibleValue:
+    """
+    This represents a pay rate compatible value.
+    """
+    name: str
+    """The value associated with the attribute, for ex, if your attribute was job_title, name would be Cashier."""
+
+    pay_rate: Optional[str] = None
```
sz2376, line 93: What's the reason why we use a str type instead of the decimal type here? Were you enforcing the field to be of 4 precision here? I feel it's ok to just take in whatever the app developer provides to us and store it as 4 precision decimal in the DB. If the app developer provides a 2 precision decimal to us that's still ok. The decimal type gives us some validations out of the box while the str type looks to be more ambiguous(I see you have added some validations in place which is addressing my concerns around the validations actually so I don't feel too strong on this then).

flux_sdk/time_and_attendance/capabilities/job_management/interface.py
```diff
@@ -15,3 +18,14 @@ def get_job_attributes(self) -> GetJobAttributesResponse:
         :return: The response containing the job attributes from the third-party system.
         """
 
+    @abstractmethod
+    def get_job_attribute_values(self) -> GetJobAttributeValuesResponse:
```
sz2376, line 22: Shall we have a GetJobAttributeValuesRequest to indicate the attributes we want to fetch? The customer might have disabled the sync for some attributes so we don't bother fetching the values for those attributes.

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: We can ask them to provide us `Decimal`, but we cannot control whether developer provides either `Decimal("23.1123")` or `Decimal(23.1123)`. If the provide the latter (by putting a float inside), it has some unintended consequence: 
```

>>> from decimal import Decimal
>>> Decimal(23.1123)
Decimal('23.112300000000001176658770418725907802581787109375')
>>> Decimal("23.1123")
Decimal('23.1123')
```

If they provide the latter, we also cannot determine how many points of precision they have provided us

So the better approach is to have them pass as string, and its very easy to do error checks because we simply try to convert to Decimal in post init, and if fail we notify the user. By forcing a string, we can also determine how many decimals they give us, and if < 4, thats fine, else we error

flux_sdk/time_and_attendance/capabilities/job_management/interface.py
dhshah-rippling: What kind of schema are you thinking?

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -10,63 +11,246 @@ class RipplingAttribute(Enum):
...
+        check_field(self, "compatible_rippling_attributes", list, required=True)
+
+        if len(self.compatible_rippling_attributes) == 0:
+            raise ValueError("Attribute error: compatible_rippling_attributes must have at least 1 value")
+
+        for attribute in self.compatible_rippling_attributes:
+            if not isinstance(attribute, RipplingAttribute):
+                raise ValueError("Attribute error: compatible_rippling_attributes must be of type RipplingAttribute")
+
+        if self.attribute_values is not None:
```
sz2376, line 170: We can also validate the attribute_values only contains types that are compatible with what's specified in the compatible_rippling_attributes. For example, we have pay_rate in the compatible_rippling_attributes but somehow there is AddressCompatibleValue in the attribute_values.

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -10,63 +11,246 @@ class RipplingAttribute(Enum):
...
+    """
+
+    requested_attribute_values: bool
+    """
+    Whether to fetch the values of the attributes or not.
+    If False, GetJobAttributesResponse will only contain the list of attributes and how they map to Rippling.
+    If True, GetJobAttributesResponse will also contain the values of the attributes in addition to their mappings.
+    """
+
+    requested_attributes: List[str] | None = None
```
sz2376, line 201:
```suggestion
    requested_attributes: list[str] | None = None
```

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -10,63 +11,246 @@ class RipplingAttribute(Enum):
...
+    compatible_rippling_attributes field.
+    """
+
+    name: str
+    """
+    The unique value associated with the attribute. Please strip any whitespaces.
+    For ex, if the attribute was "job_title", the value could be "Cashier".
+    """
+
+    associated_attribute_values: List[Union[PayRateCompatibleValue, AddressCompatibleValue]]
```
sz2376, line 100:
```suggestion
    associated_attribute_values: list[Union[PayRateCompatibleValue, AddressCompatibleValue]]
```

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -10,63 +11,246 @@ class RipplingAttribute(Enum):
...
 
-    compatible_rippling_attribute: list[RipplingAttribute]
-    """The list of Rippling attributes that are compatible with the attribute, sorted from most compatible to least."""
+    compatible_rippling_attributes: list[RipplingAttribute]
+    """
+    The non-empty list of Rippling attributes that are compatible with the attribute, 
+    sorted from most compatible to least, for ex, [PAY_RATE].
+    """
+
+    attribute_values: Optional[List[AttributeValue]] = None
```
sz2376, line 149:
```suggestion
    attribute_values: Optional[list[AttributeValue]] = None
```

sz2376 (approved):

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: See this commit - https://github.com/Rippling/flux-sdk/pull/88/commits/c7ca88ed992ef0a11dd29f80b70a0a220f3babbf

sz2376 (approved):

sihan416 (approved):

PR #92, opened 2024-10-03
[APPS-29594] add sdk for pay rate sync 

dhshah-rippling:
This is the sdk model and method used for pay rate sync for time and attendance kit.

It should be used when the pay rate information needs to be overrided , because maybe it was not provided in the `get_job_attributes` hook, or we want to do an employee level override.

Here we define a `get_employees_pay_rate_overrides` hook that should return something like the following:

```
GetEmployeesPayRateOverridesResponse(employee_pay_rate_overrides_per_attribute={
"job_title": [EmployeePayRateOverride(...),...]
}
```

Some constraints
- We only support one attribute id mapping for now -- as only one attribute maps to pay rate
- EmployeePayRateOverride cannot have repeats of `(employee_id, attribute_value_id)` -- we cannot have employee `123` working as Cashier (`456`) across different locations, as then we will not know which pay rate is accurate 
- **If the above assumption is wrong, we will have to change sdk**

image


flux_sdk/time_and_attendance/capabilities/job_management/interface.py
```diff
@@ -26,3 +27,17 @@ def get_job_attributes(self, request: GetJobAttributesRequest) -> GetJobAttribut
...
         """
+
+    @abstractmethod
+    def get_employees_pay_rate_overrides(self) -> GetEmployeesPayRateOverridesResponse:
+        """
+        A function that gets the pay rate override for each employee that requires an override.
+
+        Use this hook to fetch the pay rate override for each employee that requires it.
+
+        The pay rate override is tied to the attribute mapped to Rippling.PAY_RATE.
```
sz2376, line 38:
```suggestion
        The pay rate override is tied to the attribute mapped to RipplingAttribute.PAY_RATE.
```

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,110 @@ def __post_init__(self):
...
+            attribute_names.add(employee_attribute_name)
+
+            if employee_attribute_value.strip().lower() in employee_id_and_attribute_name_to_value[
+                (employee_id, employee_attribute_name)
+            ]:
+                raise ValueError("GetEmployeePayRateOverrideResponse error: "
+                                 "attribute_value must be unique on an (employee_id, attribute_name) basis")
+            employee_id_and_attribute_name_to_value[(employee_id, employee_attribute_name)].add(
+                employee_attribute_value.strip().lower()
+            )
```
dhshah-rippling, line 377: This is needed because otherwise we will have an employee `123` have a `job_title` as `Cashier`, potentially in more than one location.

Then we do not know which location pay rate is correct.

We are confirming with qubeyond whether this is allowed or not in their system, since we saw it on their testing data.

Realistically this should not be allowed

@sz2376

flux_sdk/time_and_attendance/capabilities/job_management/interface.py
```diff
@@ -26,3 +27,17 @@ def get_job_attributes(self, request: GetJobAttributesRequest) -> GetJobAttribut
...
+
+        If a default pay rate is given in the get_job_attributes hook, an employee may get a different one through this.
```
sz2376, line 40:
```suggestion
        The pay rate override will take precedence over the default pay rate returned from the get_job_attributes hook.
```

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,110 @@ def __post_init__(self):
                 raise ValueError(f"GetJobAttributesResponse error: duplicate name in attributes: {attribute.name}")
             ids.add(attribute.id)
             names.add(attribute.name.strip().lower())
+
+
+@dataclass(kw_only=True)
+class EmployeePayRateOverride:
+    """
+    This represents the pay rate override for a specific employee, tied to the attribute that maps to Rippling.PAY_RATE.
```
sz2376, line 281:
```suggestion
    This represents the pay rate override for a specific employee, tied to the attribute that maps to RipplingAttribute.PAY_RATE.
```

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,110 @@ def __post_init__(self):
...
+
+    For instance, assume that third party attribute job_title is mapped to Rippling.PAY_RATE.
```
sz2376, line 283:
```suggestion
    For instance, assume that third party attribute job_title is mapped to RipplingAttribute.PAY_RATE.
```

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,110 @@ def __post_init__(self):
...
+
+    Here are some example data:
+
+    EmployeePayRateOverride(employee_id="123", pay_rate="43.3943",
+    attribute_name="job_title", attribute_value="Cashier")
```
sz2376, line 288: Should this be the id of the attribute value? So in this example it would be the job_title_id

sz2376: Also the attribute_name would ideally be the attribute id, right?

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,110 @@ def __post_init__(self):
...
+
+    Employee 123 can also have a pay rate of $50.00 associated with the job title "Manager".
```
sz2376, line 290:
```suggestion
    Employee 123 can also have a pay rate of $50.00 overridden for the job title "Manager" with such record: EmployeePayRateOverride(employee_id="123", pay_rate="50.0000",
    attribute_id="job_title_attribute_id", attribute_value_id="job_title_value_id")
```

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,110 @@ def __post_init__(self):
...
+    If a pay rate is specified in the get_job_attributes hook, we will override if an employee has a specific override.
+    """
+
+    employee_id: str
+    """The unique identifier of the user in the third party system."""
+
+    attribute_name: str
+    """
+    The third party attribute name mapped to Rippling.PAY_RATE.
+    This must be the same for all employees, ie, you cannot override different attribute names for pay rate.
```
sz2376, line 301: Since this should be the same among all the overrides, do you think we can just move this field to the root level?

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,110 @@ def __post_init__(self):
...
+    """
+
+    attribute_value: str
```
sz2376, line 304: ideally it's the id of the attribute value. I guess we need to add an id field to the attribute value field here as we discussed: https://github.com/Rippling/flux-sdk/blob/main/flux_sdk/time_and_attendance/capabilities/job_management/data_models.py#L86

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,130 @@ def __post_init__(self):
...
+
+    If a pay rate is specified in the get_job_attributes hook, we will override if an employee has a specific override.
+
+    Note that an employee can have more than one override as long as the attribute_value_id is unique.
+    """
+
+    employee_id: str
+    """The unique identifier of the user in the third party system."""
+
+    attribute_value_name: str
```
sz2376, line 299: This might not be needed immediately so let's get rid of it for now?

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
```diff
@@ -272,3 +273,130 @@ def __post_init__(self):
...
+    """
+    The id of the attribute_value for this employee.
+    For ex, if the attribute_id was "job_title" and the attribute_value_name was "Cashier", the attribute_value_id
+     would be the id associated with "Cashier", for ex "123".
+    
+    Note, attribute_value_id has a 1-1 mapping with attribute_value_name for a given attribute_id.
+    For ex, (123, Cashier) is allowed, but we should not have (123, Manager) as another entry 
+    """
+
+    pay_rate: str
```
sz2376, line 315: Just a note to ourselves: when the multi-dimension pay rate override supported by T&A, we can have another field to specify more specific pay rate for the combination of job codes while the `pay_rate` field indicates the default pay rate override for the particular job code for the employee.

sz2376 (approved):

flux_sdk/time_and_attendance/capabilities/job_management/data_models.py
dhshah-rippling: removed

PR #96, opened 2024-10-17
get_break_types hook and data model definitions

sz2376:
get_break_types hook and data model definitions. All details should be found in the doc string in the code; if there is anything unclear I'd like to update the doc string as well.

### DIFF
--- a/flux_sdk/time_and_attendance/capabilities/time_entry_management/interface.py
+++ b/flux_sdk/time_and_attendance/capabilities/time_entry_management/interface.py
@@ -1,15 +1,15 @@
 from abc import ABC, abstractmethod
 
-from flux_sdk.time_and_attendance.capabilities.time_entry_management.data_models import TimeEntriesQuery, TimeEntry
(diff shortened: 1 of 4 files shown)


sihan416 (approved):

flux_sdk/time_and_attendance/capabilities/time_entry_management/interface.py
```diff
@@ -19,6 +23,19 @@ def fetch_time_entries(
...
-        """
\ No newline at end of file
+        :param query: The query filter to fetch time entries.
+        :return: A list of time entries that match the query.
+        """
+
+    @abstractmethod
+    def get_break_types(self) -> GetBreakTypesResponse:
+        """
+        A function that gets the break types defined in the third party system.
```
pshanmukrao, line 33: How will you use this ? 

Will you also have a hook to return the list of breaks taken by a worker ?

sz2376: The hook above is the one to return all the breaks. Actually let me add a break type id in the Break class.

pshanmukrao (approved):

PR #102, opened 2024-12-05
add kit

dhshah-rippling:


flux_sdk/employee_census/upload_employee_data/data_models.py
```diff
@@ -0,0 +1,28 @@
...
+class EmployeeDataRecord:
+    """
+    This represents a data record for the employee.
+    """
+    employee: Employee
+    """The object containing details about the employee."""
+
+    manager: Optional[Employee]
+    """
+    The object containing details about the employee manager, certain fields may be retrieved from here.
```
sz2376, line 21: What do you mean by "certain fields may be retrieved from here"? Let's be a little bit more concrete if we think the app developer needs to know about this; or be more concise if this is something we don't necessarily need to convey to the app developer and leave it up to them to decide how to consume the data.

flux_sdk/employee_census/upload_employee_data/interface.py
```diff
@@ -0,0 +1,16 @@
...
+from flux_sdk.flux_core.data_models import File
+
+
+class UploadEmployeeData(ABC):
+    @staticmethod
+    @abstractmethod
+    def format_employee_data(employee_data_records: list[EmployeeDataRecord]) -> File:
+        """A function that takes a list of EmployeeDataRecord and returns a formatted file.
+
+        Given a list of EmployeeDataRecord, prepare a file for upload somewhere.
```
sz2376, line 13:
```suggestion
        Given a list of EmployeeDataRecord, format them into a file based on the third party app's specs and upload it to a remote server.
```

flux_sdk/employee_census/upload_employee_data/data_models.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+
+    leave_infos: list[LeaveInfo]
+    """This object indicates the leave history of the employee."""
+
+    annual_salary: Optional[Decimal]
```
sz2376, line 27: Is a Decimal sufficient to represent the salary? Where do we specify the currency in this case?

flux_sdk/employee_census/upload_employee_data/interface.py
```diff
@@ -0,0 +1,16 @@
+from abc import ABC, abstractmethod
+
+from flux_sdk.employee_census.upload_employee_data.data_models import EmployeeDataRecord
+from flux_sdk.flux_core.data_models import File
+
+
+class UploadEmployeeData(ABC):
+    @staticmethod
+    @abstractmethod
+    def format_employee_data(employee_data_records: list[EmployeeDataRecord]) -> File:
```
sz2376, line 10: Are we sure we don't have any configuration data to be passed in? AFAIK there are some configurations we might want the customer to specify while installing the app and we want to pass them to the app developer as part of the hook invocation.

flux_sdk/employee_census/upload_employee_data/interface.py
```diff
@@ -0,0 +1,16 @@
+from abc import ABC, abstractmethod
+
+from flux_sdk.employee_census.upload_employee_data.data_models import EmployeeDataRecord
+from flux_sdk.flux_core.data_models import File
+
+
+class UploadEmployeeData(ABC):
```
sz2376, line 7:
```suggestion
class UploadEmployeeCensusData(ABC):
```

flux_sdk/employee_census/upload_employee_data/data_models.py
dhshah-rippling: Youre right, We should use monetary value as others have done. I incorporated this into the kit.

benefits kit uses this

image

flux_sdk/employee_census/upload_employee_data/data_models.py
dhshah-rippling: Will remove to make it more consise

flux_sdk/employee_census/upload_employee_data/interface.py
dhshah-rippling: I changed this to provide some configurations, specifically which environment we want to upload to and the company id.

flux_sdk/employee_census/upload_employee_census_data/interface.py
```diff
@@ -0,0 +1,21 @@
...
+    EmployeeCensusUploadSettings,
+)
+from flux_sdk.flux_core.data_models import File
+
+
+class UploadEmployeeCensusData(ABC):
+    @staticmethod
+    @abstractmethod
+    def format_employee_data(employee_census_data_records: list[EmployeeCensusDataRecord],
+                             employee_census_upload_settings: EmployeeCensusUploadSettings) -> File:
```
ujjwalshukla, line 14: maybe we should use `AppContext` here https://github.com/Rippling/flux-sdk/blob/main/flux_sdk/flux_core/data_models.py#L242C1-L267C30?

@sz2376 wdyt?

dhshah-rippling: Decided to change this to be empty settings for now

flux_sdk/employee_census/upload_employee_census_data/data_models.py
```diff
@@ -0,0 +1,31 @@
+
```
sz2376, line 1:
```suggestion
```

flux_sdk/employee_census/upload_employee_census_data/data_models.py
```diff
@@ -0,0 +1,31 @@
...
+)
+
+
+class EmployeeCensusUploadSettings:
+    """
+    This is one of the params for "format_employee_data" and has metadata about the upload.
+    """
+    pass
+
+class EmployeeCensusDataRecord:
```
sz2376, line 15: Would you like to add the `@dataclass(kw_only=True)` annotation to the class?

flux_sdk/employee_census/upload_employee_census_data/interface.py
```diff
@@ -0,0 +1,21 @@
...
+
+from flux_sdk.employee_census.upload_employee_census_data.data_models import (
+    EmployeeCensusDataRecord,
+    EmployeeCensusUploadSettings,
+)
+from flux_sdk.flux_core.data_models import File
+
+
+class UploadEmployeeCensusData(ABC):
+    @staticmethod
```
sz2376, line 11: It does not necessarily need to be static method, does it?

flux_sdk/flux_core/data_models.py
```diff
@@ -17,6 +18,21 @@ class LeaveType(Enum):
...
+    This contains employee leave of absence details corresponding to a leave type.
+    """
+    leave_type: LeaveType
+    start_date: date
+    return_date: date
+    is_paid: bool
+
+class MonetaryValue:
+    """ Object describing a value in a specific currency """
+    """ A value of currency """
```
sz2376, line 32: What does "A value of currency" mean? Could you make it more readable?

sz2376 (approved):

flux_sdk/employee_census/upload_employee_census_data/data_models.py
dhshah-rippling: Actually no in this case. If we do that we have to define everything In one call

```
EmployeeCensusDataRecord(a=xx,b=yy)
```

Instead of doing 

```
e = EmployeeCensusDataRecord()
e.a = xx
e.b = yy
```

The pension and fertility kit use approach two so I want to follow that

flux_sdk/flux_core/data_models.py
dhshah-rippling: Will remove this it shouldn't be here

dhshah-rippling: Will adjust to make it sound better

flux_sdk/employee_census/upload_employee_census_data/interface.py
dhshah-rippling: I guess I saw the same for pension and fertility so I kept it.

I guess because it doesnt require `self` static is fine

ujjwalshukla (approved):

PR #103, opened 2024-12-19
Generalized the parameters of `format_and_fetch_deduction_info`

AkshayContributes:
JIRA: https://rippling.atlassian.net/browse/BENPNP-2997

What
Generalized the parameters of format and fetch deduction info

Why: 
For Selerix the deductions file doesn’t have the RoleID field, which is essential to create the deduction Code Mapping, as a result we need to send these parameters to the ProcessEmployeesDeductions Class, which generates Deductions Object.


flux_sdk/benefits_administration/capabilities/process_employees_deductions/interface.py
```diff
@@ -18,12 +18,12 @@ class ProcessEmployeesDeductions(ABC):
     @staticmethod
     @abstractmethod
     def format_and_fetch_deduction_info(
-        stream: StringIO, deduction_code_mapping: list[ExternalDeductionCodeToRipplingCode]
+        stream: StringIO, metadata: dict[str, Any]
```
vmaniar-rippling, line 21: _**Suggestions:**_

- Use Pure Defined Models instead of dictionary or map or pair kind of data structure.
- It's reducing readability of the code.
- Concrete implementer of this abstract method will be not be able to use it properly.
- Any datatype is not type safe.
- There is no input validation in the current implementation.

AkshayContributes: For this the other alternative I have is sending a list of 
```
class EmployeeDeductionData:
role_id: str
employee_id: str
...
```

Does that sound plausible ?

vmaniar-rippling (approved):

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
```diff
@@ -23,3 +23,16 @@ class ExternalDeductionCodeToRipplingCode:
     rippling_code: str
     rippling_deduction_field: DeductionCodeField
 
+class UniqueIdTypes(Enum):
+    EMPLOYEE_ID = "EmployeeId"
+    SSN = "SSN"
+
+class EmployeeRoleData:
```
guptaaditya13, line 30: remove this

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
```diff
@@ -23,3 +23,11 @@ class ExternalDeductionCodeToRipplingCode:
     rippling_code: str
     rippling_deduction_field: DeductionCodeField
 
+class EmployeeRoleData:
+    role_id: str | None
+    unique_id: str | None
+
+class EmployeeDeductionMetaData:
```
pshanmukrao, line 30: Metadata is usually one word... D should be small case

pshanmukrao (approved):

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
```diff
@@ -23,3 +23,11 @@ class ExternalDeductionCodeToRipplingCode:
     rippling_code: str
     rippling_deduction_field: DeductionCodeField
 
+class EmployeeRoleData:
+    role_id: str | None
+    unique_id: str | None
```
guptaaditya13, line 28: instead of a generic `unique_id`
just add a single field:
`ssn: str | None`
`employee_number: str | None`

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
```diff
@@ -23,3 +23,11 @@ class ExternalDeductionCodeToRipplingCode:
     rippling_code: str
     rippling_deduction_field: DeductionCodeField
 
+class EmployeeRoleData:
+    role_id: str | None
+    employee_number: str | None
+
+class EmployeeDeductionMetadata:
+    deduction_codes: list[ExternalDeductionCodeToRipplingCode]
+    employees_role_data: list[EmployeeRoleData] | None
```
rahul-rippling, line 32: Let’s avoid using `None` otherwise, we’ll need to add checks everywhere.  
The default should be an empty list instead of `None`.

flux_sdk/benefits_administration/capabilities/process_employees_deductions/data_models.py
```diff
@@ -23,3 +23,11 @@ class ExternalDeductionCodeToRipplingCode:
     rippling_code: str
     rippling_deduction_field: DeductionCodeField
 
+class EmployeeRoleData:
+    role_id: str | None
```
rahul-rippling, line 27: Let’s avoid using None otherwise, we’ll need to add checks everywhere.
The default should be an empty list instead of None.

rahul-rippling (approved):

PR #109, opened 2025-01-09
Adds dynamic fields for specific vendors

AkshayContributes:
Problem Statement: 

We have some user defined custom fields defined for companies in the CustomFieldsV2  Model, some vendors (namely Plansource) desire the values of these field for each role, and this needs to be passed to flux-apps, so as to include them in the Census File. 

The Changes allow us to pass the Custom Field Names and their values to flux-apps.

### DIFF
--- a/flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
+++ b/flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
@@ -122,6 +122,10 @@ class EmployeeStatus:
     """ The date when the employee entered this status """
     effective_date: datetime
 
+class DynamicFields:
+    field_name: str
+    field_value: str
+
 class EmployeeHrData:
     '''
     This contains the core data about an employee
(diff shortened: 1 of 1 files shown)


flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
```diff
@@ -147,6 +151,9 @@ class EmployeeHrData:
     status: EmployeeStatus
     """ Whether or not the employee is considered eligible for benefits """
     benefits_eligibility: BenefitsEligibilityStatus
+    """ Dynamic Fields for the Vendor """
```
pshanmukrao, line 154: is it vendor specific data ? 

I think it's just custom fields that customer has provided... some vendors may be using it and some vendors may not.

flux_sdk/benefits_administration/capabilities/report_employees_hr_data/data_models.py
```diff
@@ -122,6 +122,10 @@ class EmployeeStatus:
     """ The date when the employee entered this status """
     effective_date: datetime
 
+class DynamicFields:
```
pshanmukrao, line 125: Name feels too generic. 

Can you name it in such a way that it specifies what those fields are and how those fields being provided from customer

pshanmukrao (approved):

guptaaditya13 (approved):

PR #112, opened 2025-01-17
flux-sdk for broker apps

Prawar-28:
Flux-sdk setup for broker apps capabilities - which includes uploading census file


flux_sdk/broker_app/capabilities/upload_census_data/data_models.py
```diff
@@ -0,0 +1,42 @@
+from typing import Optional
+
+
+class MemberData:
```
vmaniar-rippling, line 4: This looks like a loaded data object.
Can we please rethink on defining it in more better way?
Ex: Company Details are encapsulated into one class. employee details in one class. primary member, dependents, earning details, plan details, member health details inside member details and so on...

vmaniar-rippling: Recheck the kit name.
Should it be broker-app? As it's not an app but a kit. And kit should be reusable across apps.

Prawar-28: @vmaniar-rippling ack for changing the kit name. 
For `MemberData` , we receive census data in this flattened format itself.

vmaniar-rippling (approved):

ujjwalshukla (approved):

